PR #2, opened 2020-07-06
Initial content.

felixfontein:
GH actions needs to be adjusted, will do that once they first run :)


plugins/modules/digital_ocean_sshkey_facts.py
```diff
@@ -0,0 +1,101 @@
...
+            status_code, response.json['message']))
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=DigitalOceanHelper.digital_ocean_argument_spec(),
+        supports_check_mode=False,
+    )
+
+    module.deprecate("The 'digital_ocean_sshkey_facts' module has been deprecated, use the new 'digital_ocean_sshkey_info' module",
```
gundalow, line 94: `ERROR: plugins/modules/digital_ocean_sshkey_facts.py:0:0: missing-main-call: Did not find a call to removed_module()`

felixfontein: Already fixed, just took to long to commit since I was talking on BJ ;)

PR #16, opened 2020-08-11
digital_ocean_firewall: new module

lucasbasquerotto:
##### SUMMARY
This PR has the new module to create and update Digital Ocean Firewall rules.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
digital_ocean_firewall

##### ADDITIONAL INFORMATION
Related: https://github.com/ansible/ansible/pull/46618#issuecomment-671323757

### DIFF
--- a/plugins/modules/digital_ocean_firewall.py
+++ b/plugins/modules/digital_ocean_firewall.py
@@ -16,9 +16,9 @@ ANSIBLE_METADATA = {
 DOCUMENTATION = '''
 ---
 module: digital_ocean_firewall
-short_description: Manage cloud firewalls within Digital Ocean
+short_description: Manage cloud firewalls within DigitalOcean
 description:
-    - This module can be used to add or remove firewalls on the Digital Ocean cloud platform.
+    - This module can be used to add or remove firewalls on the DigitalOcean cloud platform.
 author:
     - Anthony Bond (@BondAnthony)
     - Lucas Basquerotto (@lucasbasquerotto)
@@ -51,7 +51,7 @@ options:
     type: list
     elements: dict
     description:
-      - Firewall rules specifically targeting inbound network traffic into Digital Ocean
+      - Firewall rules specifically targeting inbound network traffic into DigitalOcean
     required: true
     suboptions:
       protocol:
@@ -102,7 +102,7 @@ options:
     type: list
     elements: dict
     description:
-      - Firewall rules specifically targeting outbound network traffic from Digital Ocean
+      - Firewall rules specifically targeting outbound network traffic from DigitalOcean
     required: true
     suboptions:
       protocol:
@@ -325,7 +325,7 @@ class DOFirewall(object):
         response = self.rest.get("%s" % base_url)
         status_code = response.status_code
         if status_code != 200:
-            self.module.fail_json(msg="Failed to retrieve firewalls from Digital Ocean")
+            self.module.fail_json(msg="Failed to retrieve firewalls from DigitalOcean")
         return self.rest.get_paginated_data(base_url=base_url, data_key_name='firewalls')
 
     def get_firewall_by_name(self):
--- a/plugins/modules/digital_ocean_firewall.py
+++ b/plugins/modules/digital_ocean_firewall.py
@@ -52,7 +52,8 @@ options:
       protocol:
         description:
           - Network protocol to be accepted.
-        required: true
+        required: false
+        default: tcp
         choices: ['udp', 'tcp', 'icmp']
       ports:
         description:
@@ -90,7 +91,8 @@ options:
       protocol:
         description:
           - Network protocol to be accepted.
-        required: true
+        required: false
+        default: tcp
         choices: ['udp', 'tcp', 'icmp']
       ports:
         description:
@@ -263,10 +265,10 @@ from ansible_collections.community.digitalocean.plugins.module_utils.digital_oce
 from ansible.module_utils._text import to_native
 
 address_spec = dict(
-    addresses=dict(type='list', required=False),
-    droplet_ids=dict(type='list', required=False),
-    load_balancer_uids=dict(type='list', required=False),
-    tags=dict(type='list', required=False),
+    addresses=dict(type='list', elements='str', required=False),
+    droplet_ids=dict(type='list', elements='str', required=False),
+    load_balancer_uids=dict(type='list', elements='str', required=False),
+    tags=dict(type='list', elements='str', required=False),
 )
 
 inbound_spec = dict(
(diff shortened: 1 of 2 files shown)


lucasbasquerotto:
I documented a list of objects like:

```
  inbound_rules:
    description:
      - Firewall rules specifically targeting inbound network traffic into Digital Ocean
    required: true
    type: list
    suboptions:
      protocol:
      ...
```

and specified as:

`inbound_rules=dict(type='list', elements='dict', options=inbound_spec, required=True)`

but got `ERROR: plugins/modules/digital_ocean_firewall.py:0:0: doc-elements-mismatch: Argument 'inbound_rules' in argument_spec specifies elements as dict,but elements is not documented`

The same happened with outbound rules. I don't know what need to be changed.

I also defined `addresses` as `addresses=dict(type='list', required=False)`, but got:

`ERROR: plugins/modules/digital_ocean_firewall.py:0:0: parameter-list-no-elements: Argument 'addresses' in argument_spec found in inbound_rules -> sources defines type as list but elements is not defined`

Then I specified as `addresses=dict(type='list', elements='str', required=False)` (with `elements='str')`, but got:

`ERROR: plugins/modules/digital_ocean_firewall.py:0:0: doc-elements-mismatch: Argument 'addresses' in argument_spec found in inbound_rules -> sources specifies elements as str,but elements is not documented`

So I don't know exactly what is the problem here. 

If I understood correctly, if `elements `is a complex object like a dict or a list, the documentation need the `suboptions`. If it's a string the options isn't needed in the spec (and the suboptions isn't needed in the documentation), like [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/iptables.py). I looked at https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#options and don't know exatly what needs to be done. Furthermore, it documents how to make the spec, and not the documentation (based on the spec).

zetoph:
I don't have a lot of experiences with writing Ansible modules but could it be that you have to document `inbound_rules.sources.addresses.elements` explicitly?
Somewhere along those lines:
```
  inbound_rules:
    description:
      - Firewall rules specifically targeting inbound network traffic into Digital Ocean
    required: true
    type: list
    suboptions:
      protocol:
        description:
          - Network protocol to be accepted.
        required: false
        default: tcp
        choices: ['udp', 'tcp', 'icmp']
      ports:
        description:
          - The ports on which traffic will be allowed, single, range, or all
        required: true
      sources:
        description:
          - Dictionary of locations from which inbound traffic will be accepted
        required: true
        suboptions:
          addresses:
            description:
              - List of strings containing the IPv4 addresses, IPv6 addresses, IPv4 CIDRs,
                and/or IPv6 CIDRs to which the firewall will allow traffic
            required: false
            suboptions:
              elements:
                description:
                  - IPv4 or IPv6 address in CIDR notation.
                type: str
```
And then accordingly for the other 27 issues.

lucasbasquerotto:
@zetoph It seems that elements have to be defined in the docs. It seems that this is recent, and I found a file that have an example (https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/command.py) and the commit that added it in this file (and several others) is from 8 days ago (https://github.com/ansible/ansible/commit/3e9943bc5e7a9cd393757aa8100d7fed80bd316e).

So it seems that this rule is not well documented, but based on that I added `elements: [element_type]` in the docs when type is `list` or `dict`, like:

```yaml
droplet_ids:
  type: list
  elements: str
  description:
    - List of droplet ids to be assigned to the firewall
  required: false
```
and

```yaml
inbound_rules:
  type: list
  elements: dict
  description:
    - Firewall rules specifically targeting inbound network traffic into Digital Ocean
  required: true
  suboptions:
    protocol:
      ...
```

plugins/modules/digital_ocean_firewall.py
```diff
@@ -0,0 +1,525 @@
...
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: digital_ocean_firewall
+short_description: Manage cloud firewalls within Digital Ocean
+description:
+    - This module can be used to add or remove firewalls on the Digital Ocean cloud platform.
```
mamercad, line 21: Here also.

plugins/modules/digital_ocean_firewall.py
```diff
@@ -0,0 +1,525 @@
...
+    type: list
+    elements: str
+    description:
+      - List of tags to be assigned to the firewall
+    required: false
+  inbound_rules:
+    type: list
+    elements: dict
+    description:
+      - Firewall rules specifically targeting inbound network traffic into Digital Ocean
```
mamercad, line 54: And here.

plugins/modules/digital_ocean_firewall.py
```diff
@@ -0,0 +1,525 @@
...
+            elements: str
+            description:
+              - List of strings containing the names of Tags corresponding to groups of Droplets to
+                which the Firewall will allow traffic
+            required: false
+  outbound_rules:
+    type: list
+    elements: dict
+    description:
+      - Firewall rules specifically targeting outbound network traffic from Digital Ocean
```
mamercad, line 105: As well as here.

plugins/modules/digital_ocean_firewall.py
```diff
@@ -0,0 +1,525 @@
...
+        self.name = self.module.params.get('name')
+        self.baseurl = 'firewalls'
+        self.firewalls = self.get_firewalls()
+
+    def get_firewalls(self):
+        base_url = self.baseurl + "?"
+        response = self.rest.get("%s" % base_url)
+        status_code = response.status_code
+        if status_code != 200:
+            self.module.fail_json(msg="Failed to retrieve firewalls from Digital Ocean")
```
mamercad, line 328: And here please.

mamercad (changes requested):

plugins/modules/digital_ocean_firewall.py
```diff
@@ -0,0 +1,525 @@
...
+ANSIBLE_METADATA = {
+    'metadata_version': '1.1',
+    'status': ['preview'],
+    'supported_by': 'community'
+}
+
+DOCUMENTATION = '''
+---
+module: digital_ocean_firewall
+short_description: Manage cloud firewalls within Digital Ocean
```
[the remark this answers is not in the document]
lucasbasquerotto: Done

mamercad (dismissed):

mamercad (dismissed):
I've updated the name to meet DO's firewall naming requirements.

lucasbasquerotto:
@mamercad It's only now that I saw the other places you posted to fix the names, weird, because I think I didn't receive notifications after the 1st case, which is the reason I haven't fixed them previously, only now I was notificated with your comment. Thanks for the changes!

mamercad:
> @mamercad It's only now that I saw the other places you posted to fix the names, weird, because I think I didn't receive notifications after the 1st case, which is the reason I haven't fixed them previously, only now I was notificated with your comment. Thanks for the changes!

No worries, thank *you*.

mamercad:
@lucasbasquerotto Mind adding a quick [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)?

I think it'll look something like this:

```yaml
minor_changes:
  - digital_ocean_firewall - Adding a new module for DigitalOcean firewalls
```

mamercad (approved):
LGTM

mamercad:
@lucasbasquerotto Thanks very much for this contribution! 👋

PR #24, opened 2020-09-27
Run all sanity tests

felixfontein:
##### SUMMARY
Currently some sanity tests are not run at all, because the Python version is fixed to 3.6.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
CI


bmildren:
lgtm!

PR #27, opened 2020-10-10
Mass rename and redundant prefix removal

mamercad:
##### SUMMARY

The company name is "DigitalOcean", not "Digital Ocean". Also, since this code has been broken out into its own community module, the "digital_ocean_" or "digitalocean_" prefix is redundant.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

All of the things

##### ADDITIONAL INFORMATION

Yeah, I'm going to see how things look after the checks run and take it from there.


mamercad:
I'll probably let this sit until we clear some of the backlog and reintegrate...

geerlingguy:
Just leaving a note; the renaming will require any users of Ansible "community version" 2.(something) to update their older playbooks to use the new module names, and thus it's a breaking change.

I do agree we should get things named more appropriately (and less redundantly) in this collection-based world, but it will need to be part of a coordinated 2.0.0 release that will not be pulled into Ansible's community district until at least 2.11.

gundalow:
You can use `meta/runtime.yml` to redirect from the old name to the new names, meaning no impact for users.

mamercad:
> You can use `meta/runtime.yml` to redirect from the old name to the new names, meaning no impact for users.

I'll give this a look!

mamercad:
@geerlingguy Could you help nudge me in the right direction in fixing the two things that are blowing up? The errors don't quite make sense to me, it *seems* like `meta/runtime.yml` and the `DOCUMENTATION.deprecated` are lining up.

meta/runtime.yml
```diff
@@ -5,58 +5,108 @@ plugin_routing:
       deprecation:
         removal_version: 2.0.0
         warning_text: see plugin documentation for details
+      redirect: community.digitalocean.digitalocean
```
gundalow, line 8: Line 4 should be `digitalocean:`

Then add a new block for
```yaml
digital_ocean:
  redirect: community.digitalocean.digitalocean
```

gundalow: Same for ssh_key

gundalow:
```
ERROR: Found 4 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/digitalocean.py:0:0: deprecation-mismatch: "meta/runtime.yml" and DOCUMENTATION.deprecation do not agree.
ERROR: plugins/modules/digitalocean.py:0:0: invalid-documentation: DOCUMENTATION.deprecated: extra keys not allowed @ data['deprecated']. Got {'removed_in': '2.0.0', 'why': 'Updated module to remove external dependency with increased functionality.', 'alternative': 'Use M(community.digitalocean.droplet) instead.', 'removed_from_collection': 'community.digitalocean'}
ERROR: plugins/modules/sshkey_facts.py:0:0: deprecation-mismatch: "meta/runtime.yml" and DOCUMENTATION.deprecation do not agree.
ERROR: plugins/modules/sshkey_facts.py:0:0: invalid-documentation: DOCUMENTATION.deprecated: extra keys not allowed @ data['deprecated']. Got {'removed_in': '2.0.0', 'why': 'Deprecated in favour of C(_info) module.', 'alternative': 'Use M(community.digitalocean.sshkey_info) instead.', 'removed_from_collection': 'community.digitalocean'}
```


The entry in runtime.yml for the new name should be the one that includes the deprecation notice

The old name should only contain a single entry with `redirect:` to new name

mamercad:
@gundalow Ah, thanks, will press on!

meta/runtime.yml
```diff
@@ -2,61 +2,113 @@ requires_ansible: '>=2.9.10'
 plugin_routing:
   modules:
     digital_ocean:
+      redirect: community.digitalocean.digitalocean
+    digitalocean:
       deprecation:
         removal_version: 2.0.0
         warning_text: see plugin documentation for details
     digital_ocean_account_facts:
```
gundalow, line 10: This bit doesn't look right, I *think* it should be
```yaml
digital_ocean_account_facts:
  redirect: community.digitalocean.account_facts
```

Then `account_facts` has the deprecation notice

meta/runtime.yml
```diff
@@ -2,61 +2,113 @@ requires_ansible: '>=2.9.10'
...
         removal_version: 2.0.0
         warning_text: see plugin documentation for details
+      redirect: community.digitalocean.domain_facts
+    digital_ocean_domain_info:
+      redirect: community.digitalocean.domain_info
+    digital_ocean_domain:
+      redirect: community.digitalocean.domain
+    digital_ocean_droplet:
+      redirect: community.digitalocean.droplet
     digital_ocean_firewall_facts:
```
gundalow, line 35: As above

meta/runtime.yml
```diff
@@ -2,61 +2,113 @@ requires_ansible: '>=2.9.10'
...
       deprecation:
         removal_version: 2.0.0
         warning_text: see plugin documentation for details
+      redirect: community.digitalocean.firewall_facts
+    digital_ocean_firewall_info:
+      redirect: community.digitalocean.firewall_info
     digital_ocean_floating_ip_facts:
```
gundalow, line 42: As above

meta/runtime.yml
```diff
@@ -2,61 +2,113 @@ requires_ansible: '>=2.9.10'
...
       deprecation:
         removal_version: 2.0.0
         warning_text: see plugin documentation for details
+      redirect: community.digitalocean.floating_ip_facts
+    digital_ocean_floating_ip_info:
+      redirect: community.digitalocean.floating_ip_info
+    digital_ocean_floating_ip:
+      redirect: community.digitalocean.floating_ip
     digital_ocean_image_facts:
```
gundalow, line 51: as above

gundalow (commented):
I think most of the redirects where there is `deprecation` need updating

README.md
```diff
@@ -97,7 +97,7 @@ It's preferable to use content in this collection using their Fully Qualified Co
         msg: "ID is {{ my_droplet.data.droplet.id }}, IP is {{ my_droplet.data.ip_address }}"
 
     - name: Tag a resource; creating the tag if it does not exist
-      digital_ocean_tag:
+      tag:
```
felixfontein, line 100: This won't work. Either use `community.digitalocean.tag` or `digital_ocean_tag`.

README.md
```diff
@@ -9,40 +9,40 @@ Tested with the current Ansible 2.9 and 2.10 releases and the current developmen
...
-- [digital_ocean_tag](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/digital_ocean_tag_module.html) – Create and remove tag(s) to DigitalOcean resource.
-- [digital_ocean_tag_facts](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/digital_ocean_tag_facts_module.html) – Gather information about DigitalOcean tags
-- [digital_ocean_tag_info](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/digital_ocean_tag_info_module.html) – Gather information about DigitalOcean tags
-- [digital_ocean_volume_facts](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/digital_ocean_volume_facts_module.html) – Gather information about DigitalOcean volumes
-- [digital_ocean_volume_info](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/digital_ocean_volume_info_module.html) – Gather information about DigitalOcean volumes
+- [digitalocean](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/module.html) – Create/delete a droplet/SSH_key in DigitalOcean
```
felixfontein, line 12:
```suggestion
- [digitalocean](https://docs.ansible.com/ansible/2.10/collections/community/digitalocean/digitalocean_module.html) – Create/delete a droplet/SSH_key in DigitalOcean
```

README.md
```diff
@@ -119,7 +119,7 @@ If upgrading older playbooks which were built prior to Ansible 2.10 and this col
 
   tasks:
     - name: Create ssh key
-      digital_ocean_sshkey:
+      sshkey:
```
felixfontein, line 122: According to the text before this example playbook, this should **not** be changed.

tests/integration/targets/digitalocean_floating_ip/tasks/main.yml
```diff
@@ -5,7 +5,7 @@
   when: do_api_key is not defined
```
felixfontein, line 5: Regarding the target directory names: why do you keep the `digitalocean` prefix? ansible-test won't find these tests when it detect changes to the `floating_ip` module because the module is called `floating_ip`, and not `digitalocean_floating_ip`. Same for the other integration tests.

felixfontein (commented):
1. You need to document this change in a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
2. You need to add symlinks as well (next to the meta/runtime.yml entries), otherwise you're breaking backwards compatibility for Ansible 2.9 users and this change has to wait for community.digitalocean 2.0.0.
3. By changing all links in the README now, you're essentially breaking the links in README until the next Ansible 2.10.x version is released which contains a version of this collection which contains this PR. It's probably better to keep the old links until after that.

gundalow:
@mamercad Thank you for improving the DigitalOcean collection.
Do the above comments make sense?

mamercad:
> @mamercad Thank you for improving the DigitalOcean collection.
> Do the above comments make sense?

They do, I've been a bit strapped for time as of late, I will try and get to this as soon as I can.

PR #31, opened 2020-10-11
Also run CI on pushes, and nightly

### LINKED ISSUE #29: CI doesn't run on main branch, only on PRs
I think this list: https://github.com/ansible-collections/community.digitalocean/blob/main/.github/workflows/ansible-test.yml#L3 should also contain `push` as in https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml#L10.

While you're at it, you might also want to configure nightly runs (see https://github.com/ansible-collections/community.mysql/blob/main/.github/workflows/ansible-test-plugins.yml#L13-L14 for an example).

felixfontein:
Fixes #29.

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -1,6 +1,9 @@
 name: CI
 on:
-- pull_request
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 6 * * *'
 
 jobs:
   sanity:


geerlingguy (approved):

gundalow (approved):

felixfontein:
@geerlingguy @gundalow thanks a lot for reviewing!

PR #39, opened 2021-01-30
38-Add ability to resize Block Storage Volumes

### LINKED ISSUE #38: Add ability to resize Block Store Volumes
##### SUMMARY
We're missing the ability to resize Block Storage Volumes.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
digital_ocean_block_storage.py

##### ADDITIONAL INFORMATION

Basically, if we're trying to create an already-existing volume, but its size doesn't match the desired size, the volume would be resized. This would be idempotent.

DigitalOcean doesn't support shinking volumes, so that'd be a no-op

#### Example usage:

```yaml
- name: Create a volume
  digital_ocean_block_storage:
    command: create
    state: present
    volume_name: my_volume
    region: nyc1
    block_size: 15
   
- name: Resize the volume
  digital_ocean_block_storage:
    command: create
    state: present
    volume_name: my_volume
    region: nyc1
    block_size: 20  # because the desired size is larger, the volume would be enlarged
```

lalvarezguillen:
##### SUMMARY
Fixes #38 

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/plugins/modules/digital_ocean_block_storage.py
+++ b/plugins/modules/digital_ocean_block_storage.py
@@ -166,18 +166,18 @@ class DOBlockStorage(object):
         resp = self.rest.get(url)
         if resp.status_code != 200:
             raise DOBlockStorageException(resp.json['message'])
-        
+
         volumes = resp.json['volumes']
         if not volumes:
             return None
-        
+
         return volumes[0]
 
     def get_attached_droplet_ID(self, volume_name, region):
         volume = self.get_block_storage_by_name(volume_name, region)
         if not volume or not volume['droplet_ids']:
             return None
-        
+
         return volume['droplet_ids'][0]
 
     def attach_detach_block_storage(self, method, volume_name, region, droplet_id):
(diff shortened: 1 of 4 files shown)


plugins/modules/digital_ocean_block_storage.py
```diff
@@ -286,9 +326,9 @@ def main():
     try:
         handle_request(module)
     except DOBlockStorageException as e:
-        module.fail_json(msg=e.message, exception=traceback.format_exc())
```
lalvarezguillen, line 289: `Exception.message` doesn't exist on python3. But `str(Exception)` achieves the same, in all Python versions

See similar discussion in other projects for context: https://github.com/google/yapf/issues/564

plugins/modules/digital_ocean_block_storage.py
```diff
@@ -151,20 +161,24 @@ def poll_action_for_complete_status(self, action_id):
...
-            raise DOBlockStorageException(json['message'])
+        
+        return volumes[0]
+
+    def get_attached_droplet_ID(self, volume_name, region):
+        volume = self.get_block_storage_by_name(volume_name, region)
+        if not volume or not volume['droplet_ids']:
+            return None
+        
+        return volume['droplet_ids'][0]
```
lalvarezguillen, line 181: I've simply refactored this method, extracting the logic that fetches a volume by name to a separate method, so it can be reused

mamercad (approved):
Nice work, thanks for the contribution!

PR #41, opened 2021-03-04
new plugin: inventory plugin

grzs:
##### SUMMARY
New plugin: inventory plugin

##### COMPONENT NAME
community.digitalocean.digitalocean

##### ADDITIONAL INFORMATION
The plugin uses the '/v2/droplets' endpoint to acquire droplet information.
https://developers.digitalocean.com/documentation/v2/#list-all-droplets

Parameters:
* attributes: list of droplet attributes to turn to host_vars
* var_prefix: prefix of generated variable names, separated with '_' (default: do)

The host alias is generated from the name attribute, other parameters like ansible_host should be composed with jinja templates, as in the example.

This plugin imports the 'ansible.module_utils.Request' python module, because the generic helper works only with modules.
It could be possible to create a generic helper for plugins too.
I added a directory local .gitignore file to ignore emacs backup files. Should I merge it to the main .gitignore?


gundalow:
@mamercad @bmildren @Akasurde @geerlingguy I believe you are maintainers of the `community.digitalocean` collection, would you possible by able to take a look at this new inventory plugin?

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+name: digitalocean
+plugin_type: inventory
+author:
+  - Janos Gerzson (@grzs)
+short_description: DigitalOcean Inventory Plugin
+version_added: "0.04"
```
felixfontein, line 15:
```suggestion
version_added: "1.1.0"
```
This should be the next minor release version.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+description:
+  - DigitalOcean Inventory plugin.
+  - Acquires droplet list from DO api.
+  - Uses configuration file that ends with
+  - (do_hosts|digitalocean|digital_ocean).(yaml|yml)
```
felixfontein, line 20:
```suggestion
  - Uses configuration file that ends with C((do_hosts|digitalocean|digital_ocean).(yaml|yml)).
```
Generally: make sure you have complete sentences per paragraph (each list item is one paragraph), and use the proper formatting (https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation).

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+extends_documentation_fragment:
+  - community.digitalocean.digital_ocean.documentation
+  - constructed
+  - inventory_cache
+options:
+  plugin:
+    description:
+      - The name of the DigitalOcean Inventory Plugin,
+      - this should always be 'community.digitalocean.digitalocean'.
+    required: True
```
felixfontein, line 30:
```suggestion
    required: true
```
Booleans in YAML should be `true`/`false` (lowercase).

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+      Droplet attributes to add as host vars to each inventory host.
+      Check out the DO api docs for full list of attributes at
+      https://developers.digitalocean.com/documentation/v2/#list-all-droplets
+    type: list
+    default: 'id,name,networks,region,size_slug'
+  var_prefix:
+    description:
+      - Prefix of generated varible names (e.g. 'tags' -> 'do_tags')
+    type: str
+    default: 'do'
```
felixfontein, line 51: Shoudn't the prefix be `do_` to get `do_tags`? (From a user's point of view, not by looking at the code.)

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+    aliases: [ oauth_token ]
+    env:
+      - name: DO_API_TOKEN
+  attributes:
+    description: >-
+      Droplet attributes to add as host vars to each inventory host.
+      Check out the DO api docs for full list of attributes at
+      https://developers.digitalocean.com/documentation/v2/#list-all-droplets
+    type: list
+    default: 'id,name,networks,region,size_slug'
```
felixfontein, line 46: If you want a list, specify a list:
```suggestion
    default:
      - id
      - name
      - networks
      - region
      - size_slug
```

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+DOCUMENTATION = r'''
+name: digitalocean
+plugin_type: inventory
+author:
+  - Janos Gerzson (@grzs)
+short_description: DigitalOcean Inventory Plugin
+version_added: "0.04"
+description:
+  - DigitalOcean Inventory plugin.
+  - Acquires droplet list from DO api.
```
felixfontein, line 18:
```suggestion
  - Acquires droplet list from DO API.
```
Abbrevations should usually be upper-case.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,194 @@
...
+
+DOCUMENTATION = r'''
+name: digitalocean
+plugin_type: inventory
+author:
+  - Janos Gerzson (@grzs)
+short_description: DigitalOcean Inventory Plugin
+version_added: "0.04"
+description:
+  - DigitalOcean Inventory plugin.
```
felixfontein, line 17:
```suggestion
  - DigitalOcean (DO) inventory plugin.
```
Since you are using `DO` for `DigitalOcean` below, you should introduce it at some point.

plugins/inventory/digitalocean.py
grzs: Thank you for the suggestions

plugins/inventory/digitalocean.py
grzs: I was thinking of it too but wasn't sure after I saw that the keyed group generator applies an '_' as a separator by default. I don't know what users expect from an inventory plugin. If nobody vote against it within 1-2 days I apply your suggestion for this part as well.

grzs (commented):
@felixfontein thanks for the suggestions. Further comments regarding concept, code, docstring grammer or anything else are welcome.

plugins/inventory/digitalocean.py
grzs: A var_separator also could be added, but maybe it's too much...

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+name: digitalocean
+plugin_type: inventory
+author:
+  - Janos Gerzson (@grzs)
+short_description: DigitalOcean Inventory Plugin
+version_added: "1.1.0"
+description:
+  - DigitalOcean (DO) inventory plugin.
+  - Acquires droplet list from DO API.
+  - Uses configuration file that ends with C((do_hosts|digitalocean|digital_ocean).(yaml|yml)).
```
felixfontein, line 19:
```suggestion
  - Uses configuration file that ends with '(do_hosts|digitalocean|digital_ocean).(yaml|yml)'.
```
Ok, in this case the original version is better, as `C()` does not handle `(` and `)` in it well... Sorry!

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+extends_documentation_fragment:
+  - community.digitalocean.digital_ocean.documentation
+  - constructed
+  - inventory_cache
+options:
+  plugin:
+    description:
+      - The name of the DigitalOcean Inventory Plugin,
+      - this should always be 'community.digitalocean.digitalocean'.
```
felixfontein, line 28: Other places should be changed though. Also, this is part of the previous sentence, not a new paragraph, so removing the `-`:
```suggestion
        this should always be C(community.digitalocean.digitalocean).
```

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+     - DigitalOcean OAuth token.
+    required: True
+    type: str
+    aliases: [ oauth_token ]
+    env:
+      - name: DO_API_TOKEN
+  attributes:
+    description: >-
+      Droplet attributes to add as host vars to each inventory host.
+      Check out the DO api docs for full list of attributes at
```
felixfontein, line 42:
```suggestion
      Check out the DO API docs for full list of attributes at
```
I only marked some instances of problems in my first review, you need to check whether they apply to more places.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+      https://developers.digitalocean.com/documentation/v2/#list-all-droplets
```
felixfontein, line 43:
```suggestion
      U(https://developers.digitalocean.com/documentation/v2/#list-all-droplets).
```

plugins/inventory/digitalocean.py
felixfontein: Hmm, right, the keyed group generator also did that... I never understood why. In any case, feel free to do here what you want/prefer.

tadeboro (changes requested):
Thank you, @grzs, for your contribution! I only found one real issue with the changes: Python 2.7 compatibility. Since the collection claims to support Ansible >= 2.9.10, and at least one version of Ansible from that range still supports Python 2.7, we need to make sure the plugin works there.

There are a few other minor things to fix, but nothing major.

So all in all, great work!

plugins/inventory/digitalocean.py
grzs: sorry, I was lazy. Now I checked all occurrences.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+            if record['name']:
+                host_name = self.inventory.add_host(record['name'])
+            else:
+                continue
+
+            # set variables for host
+            for k in record.keys():
+                if k in attributes:
+                    self.inventory.set_variable(
+                        host_name, '%s_%s' % (var_prefix, k), record[k])
```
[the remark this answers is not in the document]
grzs: I never used items(), that's cool.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+                )
+        except Exception as error:
+            raise AnsibleParserError(error)
+
+        if result['status'] == 200:
+            payload = json.loads(result['json'])
+        else:
+            msg = 'Request failed with code {0}'.format(
+                result['code'])
+            raise AnsibleParserError(msg)
```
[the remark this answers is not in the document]
grzs: That's surely better, thank you. What do you think about making a generic API request helper for plugins like the one for modules? Or if someone would like to write another plugin, they can copy this portion from this one.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+    type: list
+    default:
+      - id
+      - name
+      - networks
+      - region
+      - size_slug
+  var_prefix:
+    description:
+      - Prefix of generated varible names (e.g. 'tags' -> 'do_tags')
```
felixfontein, line 53:
```suggestion
      - Prefix of generated varible names (e.g. C(tags) -> C(do_tags)).
```

plugins/inventory/digitalocean.py
grzs: I've chosen your suggestion.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+  plugin:
+    description:
+      - The name of the DigitalOcean Inventory Plugin,
+        this should always be 'community.digitalocean.digitalocean'.
+    required: true
+    choices: ['community.digitalocean.digitalocean']
+  api_token:
+    description:
+     - DigitalOcean OAuth token.
+    required: True
```
felixfontein, line 34:
```suggestion
    required: true
```

plugins/inventory/digitalocean.py
grzs: done

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+  ansible_host: do_networks.v4 | selectattr('type','eq','public')
+    | map(attribute='ip_address') | first
+  class: do_size.description | lower
+  distro: do_image.distribution | lower
+'''
+
+import json
+from ansible.errors import AnsibleError, AnsibleParserError
+from ansible.module_utils.urls import Request
+from ansible.plugins.inventory import BaseInventoryPlugin, Constructable, Cacheable, to_safe_group_name
```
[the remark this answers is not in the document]
grzs: done

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+    separator: ''
+compose:
+  ansible_host: do_networks.v4 | selectattr('type','eq','public')
+    | map(attribute='ip_address') | first
+  class: do_size.description | lower
+  distro: do_image.distribution | lower
+'''
+
+import json
+from ansible.errors import AnsibleError, AnsibleParserError
```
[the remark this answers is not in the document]
grzs: done

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+                     'digital_ocean.yaml', 'digital_ocean.yml')
+            ):
+                valid = True
+            else:
+                self.display.vvv(
+                    'Skipping due to inventory source file name mismatch. ' +
+                    'The file name has to end with one of the followings: ' +
+                    'do_hosts.yaml, do_hosts.yml' +
+                    'digitalocean.yaml', 'digitalocean.yml' +
+                    'digital_ocean.yaml', 'digital_ocean.yml')
```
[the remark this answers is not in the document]
grzs: done

plugins/inventory/.gitignore
```diff
@@ -0,0 +1,3 @@
+# emacs
+*~
+*#
```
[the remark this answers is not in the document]
grzs: done

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,198 @@
...
+        return valid
+
+    def parse(self, inventory, loader, path,
+              cache=True):  # Plugin interface (2)
```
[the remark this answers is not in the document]
grzs: done

felixfontein:
@grzs please don't use mentions (`@username`) in commit messages. These tend to cause cascading sequences of notifications for the person involved whenever someone pushes that commit to a branch somewhere on github (for example, when people update their `main` branches). This is very annoying to the mentioned users. Please rebase / amend the last commit to remove the `@`.

plugins/inventory/digitalocean.py
grzs: done

plugins/inventory/digitalocean.py
grzs: done

grzs:
@tadeboro is it OK for you that I appended your name to the AUTHORS list?

grzs:
@felixfontein sorry, I am not familiar with GitHub collaboration features. I've done with (hopefully) all modifications. Should I click on re-request review?

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,186 @@
...
+                results = self._cache[self.cache_key]['digitalocean']
+            except KeyError:
+                pass
+
+        if not results:
+            if self.cache_key not in self._cache:
+                self._cache[self.cache_key] = {'digitalocean': ''}
+
+        # request parameters
+        base_url = 'https://api.digitalocean.co/v2'
```
tadeboro, line 140:
```suggestion
        base_url = 'https://api.digitalocean.com/v2'
```

tadeboro (changes requested):
When I tested the plugin locally, I got back `ansible_collections.community.digitalocean.plugins.inventory.digitalocean plugin: <urlopen error [Errno -2] Name or service not known>` error. Apart from the typo in the URL, I found no other issues.

Once that one is fixed, this is ready to go as far as I am concerned.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,186 @@
...
+
+# Copyright: (c), Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+name: digitalocean
+plugin_type: inventory
```
felixfontein, line 11:
```suggestion
```
(https://github.com/ansible/ansible/issues/71795)

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,186 @@
...
+        api_token = self.get_option('api_token')
+        headers = {
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer {0}'.format(api_token)
+        }
+
+        # send request
+        req = Request(headers=headers)
+        try:
+            payload = json.load(req.get(url))
```
felixfontein, line 153: Is there some potential pagination that needs to be considered?

felixfontein (commented):
It would be nice if this plugin could have at least some basic tests (unit and/or integration).

For other collections (like community.general) this is a requirement, I have no idea what's required in this collection though.

A not complicated, but effective test would be to do some unit tests where `Request` is mocked; something between https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/inventory/test_stackpath_compute.py and https://github.com/ansible-collections/community.general/blob/main/tests/unit/plugins/inventory/test_proxmox.py maybe?

plugins/inventory/digitalocean.py
tadeboro: Ugh, yes. https://developers.digitalocean.com/documentation/v2/#links

tadeboro: I only tested this with 7 VMs and I got no pagination content back. But we definitely need to call things in a loop (and maybe even set the number of returned instances to 200 to reduce the number of requests).

grzs: ok, I'll take a look.

plugins/inventory/digitalocean.py
grzs: I thought that it's a must have

grzs: How can I run such a test? Is there a framework for this purpose?

plugins/inventory/digitalocean.py
grzs: It seems straightforward, I will implement it soon, but I am not sure if we should raise the limit to 200. Maybe there is a risk of timeout if the response is too big in some environments. The default droplet limit per account is 25 so in most of the cases it doesn't really matter. Or I can introduce a new variable with a default value of 200, so one can tune it if necessary.

grzs: I changed some variable names to follow the REST API terminology and how it is referred in DigitalOcean documentation. And it was more convenient to to append new value this way.

plugins/inventory/digitalocean.py
grzs: RTFM helped me out

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,205 @@
...
+        # send request
+        req = Request(headers=headers)
+        try:
+            self.display.vvv('Sending request to {0}'.format(url))
+            response = json.load(req.get(url))
+            key = 'droplets'
+            payload = response.get(key)
+            if not payload:
+                raise AnsibleParserError(
+                    "There is no droplet information received")
```
tadeboro, line 166: The API docs state that the `droplets` will be a list of maps, so this check is not really needed. If the DigitalOcean decides to break the compatibility, the kind of exception we get does not matter much, so we can just let the plugin error-out with a key error and be done with it.

```suggestion
            payload = response['droplets']
```

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,205 @@
...
+            if response.get('links'):
+                while response.get('links').get('pages').get('next'):
+                    url = response.get('links').get('pages').get('next')
+                    self.display.vvv('Sending request to {0}'.format(url))
+                    response = json.load(req.get(url))
+                    payload += response.get(key)
```
tadeboro, line 172: This part is a bit awkward and convoluted. Maybe we should simplify it into something like this:

```suggestion
        # request parameters
        api_token = self.get_option('api_token')
        headers = {
            'Content-Type': 'application/json',
            'Authorization': 'Bearer {0}'.format(api_token)
        }

        # send request
        req = Request(headers=headers)
        url = 'https://api.digitalocean.com/v2/droplets?per_page' + str(pagination)
        payload = []
        try:
            while url:
                self.display.vvv('Sending request to {0}'.format(url))
                response = json.load(req.get(url))
                payload.extend(response['droplets'])
                url = response.get('links', {}).get('pages', {}).get('next')
```

plugins/inventory/digitalocean.py
grzs: ok, let it be like that. After examining the unit tests maybe I became a little bit paranoid.

plugins/inventory/digitalocean.py
grzs: wow, that's elegant. After I committed this I started to think of a 'do while' like approach, but that's certainly better. Thank you! I still don't understand the empty maps in the last line, but I believe that's good. It's above my skills.

grzs: One thing I rolled back (by accident the pagination variable went to trash with the other garbage). The url line was too long (I don't know how strict Ansible follows the pep8 recommendations). So I decided to break the url to logical parts, because it makes easier to adopt it to other cases or move the url builder to a separate class like the DigitalOceanHelper class in module_utils. What do you think?

tadeboro: If the refactoring ever happens, most of the code related to DO API interaction will be removed and replaced by a generalized HTTP API client from [https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/module_utils/digital_ocean.py). So whatever we do here is good enough.

BTW, that utility module is quite good, but in the current form, it cannot be used from non-module code. With a little help, it could be transformed into something really nice.

As for the line length limit, Ansible in general standardized on 160 characters per line, and this limit is enforced by the sanity tests.

tadeboro (changes requested):
This code is ready to be merged once the changelog fragment is added. Unit and/or integration tests are, as far as I am concerned, in the nice-to-have category here because:

 1. unit tests will require a non-trivial amount of mocking for a minimal gain in confidence,
 2. integration tests will either need a dummy API server or credentials for the real thing.

All in all, I like the current state of things. Thank you, @grzs, for persevering!

grzs:
Thank you @tadeboro for all your help. I already wrote some unit tests, I can push the commit maybe today. For that I had to refactor the code. Do you want to look at it before the merge?

grzs:
Here is the new version with basic unit testing. However, I encountered several new issues:My problem is that this version doesn't pass the verify_config on my Debian10 container with python3.9, but the tests are successful on my Mac OS host with python3.8. I

plugins/inventory/digitalocean.py
grzs: You convinced me, I simplified the url line

tests/unit/plugins/inventory/test_digitalocean.py
```diff
@@ -0,0 +1,178 @@
...
+    config = {
+        'plugin': 'community.digitalocean.digitalocean',
+        'api_token': 'xxxxxxxxxxxxxxxxxxxx',
+        'attributes': ['id', 'locked', 'size'],
+        'var_prefix': 'foo-',
+        'pagination': 25
+    }
+    with pytest.raises(AnsibleParserError) as error_message:
+        inventory._validate_config(config)
+        assert "var_prefix contains invalid characters" in error_message
```
tadeboro, line 47: Nitpick:

```suggestion
    with pytest.raises(AnsibleParserError, match='var_prefix contains invalid characters'):
        inventory._validate_config(config)
```

tests/unit/plugins/inventory/test_digitalocean.py
```diff
@@ -0,0 +1,178 @@
...
+__metaclass__ = type
+
+import pytest
+
+from ansible.errors import AnsibleParserError
+from ansible.inventory.data import InventoryData
+from ansible_collections.community.digitalocean.plugins.inventory.digitalocean import InventoryModule
+
+
+@pytest.fixture(scope="module")
```
tadeboro, line 15: Is there a reason for scoping this fixture to a test module? This feels unsafe since inventory plugins are not "pure" and do have mutable internal states. I would suggest that we leave out the scope parameter and just use the default function scope that will deliver a fresh instance on each test function invocation.

tests/unit/plugins/inventory/test_digitalocean.py
```diff
@@ -0,0 +1,178 @@
...
+    config = {
+        'plugin': 'community.digitalocean.digitalocean',
+        'api_token': 'xxxxxxxxxxxxxxxxxxxx',
+        'attributes': ['id', 'locked', 'foo', 'size'],
+        'var_prefix': 'foo_',
+        'pagination': 25
+    }
+    with pytest.raises(AnsibleParserError) as error_message:
+        inventory._validate_config(config)
+        assert "invalid attribute: foo" in error_message
```
tadeboro, line 36: Nitpick:

```suggestion
    with pytest.raises(AnsibleParserError, match='invalid attribute: foo'):
        inventory._validate_config(config)
```

tests/unit/plugins/inventory/test_digitalocean.py
```diff
@@ -0,0 +1,178 @@
...
+    config = {
+        'plugin': 'community.digitalocean.digitalocean',
+        'api_token': 'xxxxxxxxxxxxxxxxxxxx',
+        'attributes': ['id', 'locked', 'size'],
+        'var_prefix': 'foo_',
+        'pagination': 205
+    }
+    with pytest.raises(AnsibleParserError) as error_message:
+        inventory._validate_config(config)
+        assert "pagination value is greater than the maximum" in error_message
```
tadeboro, line 58: Nitpick:

```suggestion
    with pytest.raises(AnsibleParserError, match='pagination value is greater than the maximum') as error_message:
        inventory._validate_config(config)
```

tests/unit/plugins/inventory/test_digitalocean.py
```diff
@@ -0,0 +1,178 @@
...
+    r = InventoryModule()
+    r.inventory = InventoryData()
+    return r
+
+
+def test_verify_file_bad_config(inventory):
+    assert inventory.verify_file('digitalocean_foobar.yml') is False
+
+
+def test_validate_config(inventory):
```
tadeboro, line 26: Each test function should only have one set of assert statements. I would suggest that you split the `test_validate_config` into three separate functions named `test_validate_config_invalid_attribute`, `test_validate_config_invalid_prefix`, and `test_validate_config_invalid_pagination`. You may also want to consider adding a test on a valid data.

tadeboro (commented):
It feels like the latest batch of changes made code less generic and future-proof by introducing constraints we do not have control over (what fields API returns, for example). But since I do not have the permissions to merge things, I will leave the final decision to the person who does have them.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,238 @@
...
+        do_fields = ['id', 'name', 'memory', 'vcpus', 'disk', 'locked', 'created_at', 'status', 'backup_ids', 'snapshot_ids', 'features',
+                     'region', 'image', 'size', 'size_slug', 'networks', 'kernel', 'next_backup_window', 'tags', 'volume_ids', 'vpc_uuid']
+
+        try:
+            attributes = config['attributes']
+            for a in attributes:
+                if a not in do_fields:
+                    raise AnsibleParserError("invalid attribute: " + a)
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
felixfontein: I agree it's probably best to not validate this (and same for `pagination` below). This will be a constant maintenance burden if anything ever changes on DO's side.

Validating against an actual response would be possible, but that would make everything more complex (also there's the case when the response contains no entries - do we skip validation then?).

tests/unit/plugins/inventory/test_digitalocean.py
felixfontein: I guess this comes from some of the existing tests (I've seen it there as well). I agree that not specifying `scope="module"` is better.

tests/unit/plugins/inventory/test_digitalocean.py
felixfontein: Alternatively de-dent the `assert`. The current way, the assertion won't be checked.

tests/unit/plugins/inventory/test_digitalocean.py
felixfontein: I'd personally remove the validation code from the plugin, and thus also the tests for them :)

grzs: Regarding pagination and attributes I am happily do so, but is it safe to leave the user to set whatever var_prefix (e.g. 'my do-')?

felixfontein: `my do-` can also be used, for example `hostvars[inventory_hostname]['my do-status']`. That's not recommended, but possible.

plugins/inventory/digitalocean.py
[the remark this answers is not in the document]
grzs: I had that feeling too, but wanted to try all possibilities I can. Let's throw it out!

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,238 @@
...
+                raise AnsibleParserError("var_prefix contains invalid characters")
+        except KeyError:
+            pass
+
+        try:
+            pagination = config['pagination']
+            if pagination > 200:
+                raise AnsibleParserError("pagination value is greater than the maximum (200)")
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
grzs: ditto

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,238 @@
...
+from ansible.plugins.inventory import BaseInventoryPlugin, Constructable, Cacheable
+
+
+class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
+
+    NAME = 'community.digitalocean.digitalocean'
+
+    def _validate_config(self, config):
+        if config['plugin'] != 'community.digitalocean.digitalocean':
+            raise AnsibleParserError("plugin doesn't match this plugin")
```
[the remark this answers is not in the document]
grzs: ok

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,238 @@
...
+
+        try:
+            attributes = config['attributes']
+            for a in attributes:
+                if a not in do_fields:
+                    raise AnsibleParserError("invalid attribute: " + a)
+        except KeyError:
+            pass
+
+        try:
```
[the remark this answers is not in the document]
grzs: tried, works, changed

tests/unit/plugins/inventory/test_digitalocean.py
grzs: er... yes, I copied what I found without knowing what it does... deleted and still works :)

tests/unit/plugins/inventory/test_digitalocean.py
grzs: deleted pagination test

tests/unit/plugins/inventory/test_digitalocean.py
grzs: what to assert if no error message? 😕

tests/unit/plugins/inventory/test_digitalocean.py
grzs: see below

tests/unit/plugins/inventory/test_digitalocean.py
grzs: I believe you and not bond to it emotionally. If it's ok for @tadeboro as well, I wipe out the whole section.

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,215 @@
...
+      - Has to start with a letter,
+        the only valid special character is underscore (C(_)).
+    type: str
+    default: 'do_'
+  pagination:
+    description:
+      - Maximum droplet objects per response page.
+      - If the number of droplets related to the account exceeds this value,
+        the query will be broken to multiple requests (pages).
+      - This value cannot be greater than 200.
```
felixfontein, line 64: How about:
```suggestion
      - DigitalOcean currently allows a maximum of 200.
```

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,215 @@
...
+
+        # send request(s)
+        self.req = Request(headers=headers)
+        payload = []
+        try:
+            while url:
+                self.display.vvv('Sending request to {0}'.format(url))
+                response = json.load(self.req.get(url))
+                payload.extend(response['droplets'])
+                url = response.get('links', {}).get('pages', {}).get('next')
```
felixfontein, line 156: Did someone test whether that works (for example by setting pagination to 1, or whatever is the smallest value DO allows)?

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,215 @@
...
+        except ValueError:
+            raise AnsibleParserError("something went wrong with json loading")
```
felixfontein, line 158:
```suggestion
            raise AnsibleParserError("something went wrong with JSON loading")
```

plugins/inventory/digitalocean.py
```diff
@@ -0,0 +1,215 @@
...
+from ansible.plugins.inventory import BaseInventoryPlugin, Constructable, Cacheable
+
+
+class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
+
+    NAME = 'community.digitalocean.digitalocean'
+
+    def _validate_config(self, config):
+        if 'var_prefix' in config:
+            if not re.match('^[a-z][a-z_]*$', config['var_prefix']):
```
felixfontein, line 113: I'd probably leave this away. If you really want to kee pit, you should also allow digits after the first letter.

plugins/inventory/digitalocean.py
grzs: done

plugins/inventory/digitalocean.py
grzs: I tested it with different values between 2 and 25 and worked as expected.

plugins/inventory/digitalocean.py
grzs: done

plugins/inventory/digitalocean.py
grzs: I deleted the _verify_config block. So no more questions left I think.

felixfontein (approved):

tests/unit/plugins/inventory/test_digitalocean.py
grzs: it has been deleted

grzs:
@felixfontein and @tadeboro , thank you for the patient guidance. Without your help I could never get such a quality code. I generated the changelog with antsibull-changelog, I don't know if it's good or not. So, what's next? Should I call for a collection admin to review the changes and merge?

CHANGELOG.rst
```diff
@@ -5,6 +5,17 @@ Community DigitalOcean Release Notes
 .. contents:: Topics
 
 
+v1.1.0
```
felixfontein, line 8: This change should NOT have been in this PR, and neither the change for changelog.yaml.

These files must ONLY be updated on release.

grzs:
@felixfontein I am sorry, I thought this is my task to prepare the
changelog. I didn't find good examples. Now the PR has been merged. Should
I revert the last commit? Can you please point to a documentation of the
release process?

Thank you @mamercad for expecting my PR. Maybe a little adjustment needed
in the changelog.


Felix Fontein ***@***.***> ezt írta (időpont: 2021. márc.
25., Cs 22:28):

mpontillo:
Don't worry about it @grzs - it was our misunderstanding that caused the error, so I've already reverted the modification to the changelog. Thanks for your contribution!

mpontillo:
For future reference, [here is the documentation for changelog fragments](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment); these can be included in PRs without a problem.

grzs:
Thank you @mpontillo!

PR #44, opened 2021-03-17
Fixes rel inventory script --host option: don't throw ValueError on invalid droplet ID, return JSON conformant with specification.

sgpinkus:
##### SUMMARY

Two minor fixes related to the `--host` option to the inventory script:

1. Currently if a non integer value is passed the script throws this back at you:

```
python ./hosts/digital_ocean.py --host abc
Traceback (most recent call last):
  File "./hosts/digital_ocean.py", line 541, in <module>
    DigitalOceanInventory()
  File "./hosts/digital_ocean.py", line 293, in __init__
    json_data = self.load_droplet_variables_for_host()
  File "./hosts/digital_ocean.py", line 484, in load_droplet_variables_for_host
    host = int(self.args.host)
ValueError: invalid literal for int() with base 10: 'abc'
```
This PR fixes this by declaring that the `--host` arg should be an int in `parseargs`, which makes it a bit clearer what the HOST arg is actually supposed to be:

```
python ./hosts/digital_ocean.py --host abc
usage: digital_ocean.py [-h] [--list] [--host HOST] [--all] [--droplets]
                        [--regions] [--images] [--sizes] [--ssh-keys]
                        [--domains] [--tags] [--pretty]
                        [--cache-path CACHE_PATH]
                        [--cache-max_age CACHE_MAX_AGE] [--force-cache]
                        [--refresh-cache] [--env] [--api-token API_TOKEN]
digital_ocean.py: error: argument --host: invalid int value: 'abc'
```

2. According to the [spec](https://docs.ansible.com/ansible/latest/dev_guide/developing_inventory.html#inventory-script-conventions), with `--host` option, inventory scripts are supposed to return one of two things: an object with vars at the *top level* (currently they are nested like `{ droplet: { ... } }`, PR removes the `droplet` nesting), or an empty object `{}`. Spec doesn't say specifically what should be returned if host does not exist, but in lack of further spec I've presumed it should be `{}`. This PR tweaks the output of `--host` to be consistent with that. It doesn't matter much since the script returns `_meta` and `--host` should not be used by Ansible, but still think given a either/or choice it's best to be consistent with spec.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

scripts/inventory/digital_ocean.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/44-fixes-inv-script-host-option.yaml
+++ b/changelogs/fragments/44-fixes-inv-script-host-option.yaml
@@ -1,0 +1,3 @@
+bugfixes:
+  - Fail cleaner on invalid ``HOST`` arg to inventory script ``--host`` option (https://github.com/ansible-collections/community.digitalocean/pull/44).
+  - Return JSON consistent with spec from inventory script with ``--host`` (https://github.com/ansible-collections/community.digitalocean/pull/44).
--- a/changelogs/fragments/44-fixes-inv-script-host-option.yaml
+++ b/changelogs/fragments/44-fixes-inv-script-host-option.yaml
@@ -1,3 +1,3 @@
 bugfixes:
   - digital_ocean inventory script - fail cleaner on invalid ``HOST`` argument to ``--host`` option (https://github.com/ansible-collections/community.digitalocean/pull/44).
-  - Return JSON consistent with spec from inventory script with ``--host`` (https://github.com/ansible-collections/community.digitalocean/pull/44).
+  - digital_ocean inventory script - return JSON consistent with specification with ``--host`` (https://github.com/ansible-collections/community.digitalocean/pull/44).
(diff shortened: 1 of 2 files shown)


sgpinkus:
P.S. doc fragment to come (need the PR number to complete the fragment ...).

changelogs/fragments/44-fixes-inv-script-host-option.yaml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - Fail cleaner on invalid ``HOST`` arg to inventory script ``--host`` option (https://github.com/ansible-collections/community.digitalocean/pull/44).
```
felixfontein, line 2:
```suggestion
  - digital_ocean inventory script - fail cleaner on invalid ``HOST`` argument to ``--host`` option (https://github.com/ansible-collections/community.digitalocean/pull/44).
```

changelogs/fragments/44-fixes-inv-script-host-option.yaml
```diff
@@ -0,0 +1,3 @@
...
+  - Return JSON consistent with spec from inventory script with ``--host`` (https://github.com/ansible-collections/community.digitalocean/pull/44).
```
felixfontein, line 3:
```suggestion
  - digital_ocean inventory script - return JSON consistent with specification with ``--host`` (https://github.com/ansible-collections/community.digitalocean/pull/44).
```

sgpinkus: @felixfontein Of course! Forgot about this, thanks!

mamercad (approved):
LGTM, thanks for fixing this.

PR #45, opened 2021-03-18
Implement documented but unimplemented `use_private_network` INI config option.

sgpinkus:
##### SUMMARY

`use_private_network` is documented as a inventory script INI config option. This options is present in code but it's not hooked up to anything. This PR hooks it up in a manner consistent with the description of how it should work that exist in code:

> Use the private network IP address instead of the public when available.

In addition, code docs states `do_ip_address` and `do_private_ip_address` variables are registered by the inventory script. However, they weren't. This PR registers these fields as they are used in implementing `use_private_network`.

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

scripts/inventory/digital_ocean.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/45-fix-use_private_network.yaml
+++ b/changelogs/fragments/45-fix-use_private_network.yaml
@@ -1,4 +1,4 @@
 bugfixes:
-  - Implement unimplemented ``use_private_network`` option to digital_ocean.py
+  - digital_ocean inventory script - implement unimplemented ``use_private_network`` option
     inventory script, and register missing ``do_ip_address``, ``do_private_ip_address``
     host vars (https://github.com/ansible-collections/community.digitalocean/pull/45/files).
--- a/changelogs/fragments/45-fix-use_private_network.yaml
+++ b/changelogs/fragments/45-fix-use_private_network.yaml
@@ -1,0 +1,4 @@
+bugfixes:
+  - Implement unimplemented ``use_private_network`` option to digital_ocean.py
+    inventory script, and register missing ``do_ip_address``, ``do_private_ip_address``
+    host vars (https://github.com/ansible-collections/community.digitalocean/pull/45/files).
--- a/changelogs/fragments/45-fix-use_private_network.yaml
+++ b/changelogs/fragments/45-fix-use_private_network.yaml
@@ -1,4 +1,4 @@
 bugfixes:
   - digital_ocean inventory script - implement unimplemented ``use_private_network`` option
-    inventory script, and register missing ``do_ip_address``, ``do_private_ip_address``
+    and register missing ``do_ip_address``, ``do_private_ip_address``
     host vars (https://github.com/ansible-collections/community.digitalocean/pull/45/files).
--- a/scripts/inventory/digital_ocean.py
+++ b/scripts/inventory/digital_ocean.py
@@ -445,15 +445,19 @@ class DigitalOceanInventory(object):
         for droplet in self.data['droplets']:
             for net in droplet['networks']['v4']:
                 if net['type'] == 'public':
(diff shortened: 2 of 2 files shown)


changelogs/fragments/45-fix-use_private_network.yaml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - Implement unimplemented ``use_private_network`` option to digital_ocean.py
```
felixfontein, line 2:
```suggestion
  - digital_ocean inventory script - implement unimplemented ``use_private_network`` option
```

changelogs/fragments/45-fix-use_private_network.yaml
```diff
@@ -0,0 +1,4 @@
...
+    inventory script, and register missing ``do_ip_address``, ``do_private_ip_address``
```
felixfontein, line 3:
```suggestion
    and register missing ``do_ip_address``, ``do_private_ip_address``
```

mamercad:
Is my interpretation that if the Droplet has both Public and Private addresses, the Public will trump the Private as far as this goes? And, if the Droplet has only Private networking (this would be broken before this PR) it'd be grouped by the Private address?

sgpinkus:
@mamercad  Currently DO droplets *always* have public networking. You can't un-assign a droplet's public IP address (which is kind of strange, there is an old open feature request to allow it, but doesn't seem like a priority). If you set `private_networking: true`  on droplet creation you get private networking too. Setting the inventory script `use_private_networking: True` will try to use the private network address as the host identifier, `inventory_hostname`, but silently fallback to using public IP address if the private IP address is not available.

There are variants of course: bail if private address is unavailable, bail if public IP address is not present either to future proof, better handling of what happens if no address is available. I just did minimum to get this documented but unimplemented feature working (I'm using it currently) inline with the code that's already there. It's not a perfect solution, private IP is just slightly better identifier in some use cases and easy to implement given existing code.

It's not a perfect solution because a droplet actually has at least four plausible identifiers: name, id, public IP, private IP. Ideally we should be able to identify hosts by name. This allows you to more easily target the host in other inventories (for example your static project inventory), since name is known apriori and human readable, but IPs aren't.  However, `name` is slightly problematic because the DO backend does not require names to be unique.  The current work around is to target groups not hosts.

mamercad:
> @mamercad Currently DO droplets _always_ have public networking. You can't un-assign a droplet's public IP address (which is kind of strange. There is an old open feature request to allow it, but doesn't seem like a priority). If you set `private_networking: true` on droplet creation you get private networking too. Setting the inventory script `use_private_networking: True` will try to use the private network address as the host identifier, `inventory_hostname`, but silently fallback to using public IP address if the private IP address is not available.
> 
> There are variants of course: bail if private address is unavailable, bail if public IP address is not present either to future proof, better handling of what happens if no address is available. I just did minimum to get this documented but unimplemented feature working (I'm using it currently) inline with the code that's already there. It's not a perfect solution, private IP is just slightly better identifier in some use cases and easy to implement given existing code.
> 
> It's not a perfect solution because a droplet actually has at least four plausible identifiers: name, id, public IP, private IP. Ideally we should be able to identify hosts by name. This allows you to more easily target the host in other inventories (for example your static project inventory), since name is known apriori and human readable, but IPs aren't. However, `name` is slightly problematic because the DO backend does not require names to be unique. The current work around is to target groups not hosts in your other inventory sources.

Ah, thanks for clearing that up for me, and thanks for shoring up this gap.

mamercad (approved):
LGTM

PR #53, opened 2021-04-10
Add digital_ocean_droplet_info module

### LINKED ISSUE #34: RFE: List current droplets in DO account
##### SUMMARY

First, thanks for this collection. Trying to get into it a bit :) It does not appear possible to retrieve a list of droplets. I'm wondering if I've missed something, or if this is a feature gap, or something that's in progress. It seems like a fairly basic operation that should be possible, which is why I doubt myself :)

DO API: https://developers.digitalocean.com/documentation/v2/#list-all-droplets

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

digital_ocean_droplet

##### ADDITIONAL INFORMATION

Essentially, getting a list of droplets would allow operations to be run on all current, or a subset by tag/etc, of droplets, without needing to keep an inventory of droplet IDs. 

Forgive me if this isn't a good idea - still getting used to Ansible :)

```yaml

```

N/A

tylerauerbeck:
##### SUMMARY

This PR adds the digital_ocean_droplet_info module which allows you to return info on:
- all droplets within an account
- search for droplet info by a specific id
- search for droplet info by a specific name

Fixes #34 

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
digital_ocean_droplet_info

##### ADDITIONAL INFORMATION
N/A


tadeboro (changes requested):
Thank you for your contribution, @tylerauerbeck! The general idea behind the new module is fine, but the implementation needs a bit more polish before it could be considered robust and user-friendly. I highlighted the problematic part in the code and, if at all possible, also added a suggested fix or some hints about the improved implementation.

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,190 @@
...
+      response = rest.get("droplets")
+
+      if response.status_code != 200:
+        module.fail_json(msg="Failed to fetch 'droplets' information due to error : %s" % response.json['message'])
+
+      else:
+        for droplet in response.json["droplets"]:
+          if droplet["name"] == module.params["name"]:
+            data = droplet
+            break
```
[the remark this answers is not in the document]
mamercad: Droplet names aren't unique:

```
❯ doctl compute droplet list | awk '{print $1" "$2}'
ID Name
240796291 testdrop1
240796296 testdrop1
```

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,190 @@
...
+            },
+            "size_slug": "s-1vcpu-1gb",
+            "snapshot_ids": [],
+            "status": "active",
+            "tags": ["tag1"],
+            "vcpus": 1,
+            "volume_ids": [],
+            "vpc_uuid": "1f5e24b6-8c1d-4ace-bd8f-430096ba42ca"
+        }
+    ]
```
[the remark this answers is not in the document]
tylerauerbeck: Fixed this and went with the block style you mention

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,190 @@
...
+      if response.status_code != 200:
+        module.fail_json(msg="Failed to fetch 'droplets' information due to error : %s" % response.json['message'])
+
+      else:
+        for droplet in response.json["droplets"]:
+          if droplet["name"] == module.params["name"]:
+            data = droplet
+            break
+          else:
+            data = {}
```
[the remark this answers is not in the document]
tylerauerbeck: Agreed. This is much cleaner. Fixed.

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,190 @@
...
+    argument_spec = DigitalOceanHelper.digital_ocean_argument_spec()
+    argument_spec.update(
+            name=dict(type='str', required=False, default=None),
+            id=dict(type='str', required=False, default=None)
+    )
+    module = AnsibleModule(argument_spec=argument_spec)
+    try:
+        core(module)
+    except Exception as e:
+        module.fail_json(msg=to_native(e), exception=format_exc())
```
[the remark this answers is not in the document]
tylerauerbeck: Agreed. Was originally just going off of some of the existing examples. But run is much more understandable. Fixed.

tylerauerbeck:
@tadeboro Thanks for all of the feedback! Incorporated the changes you suggested. Working on fixing the failing tests now and then this should be ready.

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,245 @@
...
+      vcpus: 1
+      volume_ids: []
+      vpc_uuid: 123-abc-567a
+  type: list
+'''
+
+from traceback import format_exc
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.digitalocean.plugins.module_utils.digital_ocean import DigitalOceanHelper
+from ansible.module_utils._text import to_native
```
tadeboro, line 206: This import is not needed anymore.

```suggestion
```

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,245 @@
...
+        data = response.json["droplets"]
+
+    module.exit_json(changed=False, data=data)
+
+
+def main():
+    argument_spec = DigitalOceanHelper.digital_ocean_argument_spec()
+    argument_spec.update(
+        name=dict(type='str', required=False, default=None),
+        id=dict(type='str', required=False, default=None)
```
tadeboro, line 238: The module code treats these two arguments as mutually exclusive (it assumes that at most one will be set), but the parameter validation does not enforce that. So we either need to mark those two as mutually exclusive (and of course also document that fact in the documentation block) or change the run method to be able to handle the situation where both are present.

See https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#dependencies-between-module-options for more information on how to specify parameter dependencies.

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,245 @@
...
+    rest = DigitalOceanHelper(module)
+
+    if module.params["id"]:
+        path = "droplets/" + module.params["id"]
+        return_codes = (200, 404)
+    else:
+        path = "droplets"
+        return_codes = (200)
+
+    response = rest.get(path)
```
tadeboro, line 219: In my original review, I forgot that the droplet API is paginated. This line of code will fail to retrieve complete data if the user has more than 20 droplets (at least I think the default page size is 20).

So we actually need to use  https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/module_utils/digital_ocean.py#L98 here when the id is not set.

tadeboro (changes requested):
I did the second review, and there are still a few issues we need to solve (I added comments to the relevant parts of the code).

But there is also one a bit higher-level issue: how should the module behave when the requested droplet is missing? Right now, the module parameters serve as filters, which means that the module invocation will never fail. And while I see this as a correct behavior for this info module, other info modules behave differently.

For example, https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_certificate_info.py#L87, https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_domain_info.py#L99, https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_load_balancer_info.py#L89, and https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_tag_info.py#L89 all fail if the resource we wanted to retrieve information about is missing, but do not fail if we want to list all resources. So we need to decide what behavior do we want and then enforce it across all modules.

Right now, I am leaning towards making this module behave like the rest of the info modules and fail if the id is specified and the droplet is missing. I would much rather see if we could change the other modules, but that would potentially mean breaking existing playbooks and would thus require a version 2.0.0 of the DI collection.

@mamercad Any thoughts on how we should proceed here?

tylerauerbeck:
@tadeboro Looks like everything is in good shape. Let me know if anything else needs done.

mamercad:
> @tadeboro Looks like everything is in good shape. Let me know if anything else needs done.

Thanks for writing this, mind updating the `README.md`?

mamercad:
> > @tadeboro Looks like everything is in good shape. Let me know if anything else needs done.
> 
> Thanks for writing this, mind updating the `README.md`?

There are a few things left as well (there are a few unresolved conversations).

mamercad:
> I did the second review, and there are still a few issues we need to solve (I added comments to the relevant parts of the code).
> 
> But there is also one a bit higher-level issue: how should the module behave when the requested droplet is missing? Right now, the module parameters serve as filters, which means that the module invocation will never fail. And while I see this as a correct behavior for this info module, other info modules behave differently.
> 
> For example,
> 
> https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_certificate_info.py#L87
> 
> ,
> https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_domain_info.py#L99
> 
> ,
> https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_load_balancer_info.py#L89
> 
> , and
> https://github.com/ansible-collections/community.digitalocean/blob/f93b52c7cd0ed78b7864a7f0d650143e43d59e3e/plugins/modules/digital_ocean_tag_info.py#L89
> 
> all fail if the resource we wanted to retrieve information about is missing, but do not fail if we want to list all resources. So we need to decide what behavior do we want and then enforce it across all modules.
> Right now, I am leaning towards making this module behave like the rest of the info modules and fail if the id is specified and the droplet is missing. I would much rather see if we could change the other modules, but that would potentially mean breaking existing playbooks and would thus require a version 2.0.0 of the DI collection.
> 
> @mamercad Any thoughts on how we should proceed here?

I echo the sentiment that for the time being, before 2.0.0, we should strive for consistency of the "info" modules. So, this one should fail in the same way as the other info modules if the Droplet isn't found.

tylerauerbeck:
Sorry for the delay here. Working to get these last issues fixed up. Looking to push the fixes here shortly.

mamercad (changes requested):
Requesting changes (in the comments).

plugins/modules/digital_ocean_droplet_info.py
tylerauerbeck: Fixed.

plugins/modules/digital_ocean_droplet_info.py
tylerauerbeck: Fixed.

plugins/modules/digital_ocean_droplet_info.py
tylerauerbeck: Fixed.

tylerauerbeck:
@tadeboro @mamercad Fixed the PR comments here and the default behavior (failing if name or id don't return matches). Will work on the integration test and README today to wrap this up.

mamercad:
@tylerauerbeck I've triggered the tests (I believe they'll run next time you push, automatically).

tylerauerbeck:
@mamercad Added the integration test and updated the README. Looks like you need to approve each run of the workflows now though.

mamercad:
> @mamercad Added the integration test and updated the README. Looks like you need to approve each run of the workflows now though.

Ah, okay, I'll keep an eye out, thank you.

mamercad:
@tylerauerbeck After fixing up the sanity tests, you'll need to add an [integration alias](https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/integration-aliases.html) as well like [this](https://github.com/ansible-collections/community.digitalocean/blob/main/tests/integration/targets/digital_ocean_account_info/aliases).

tylerauerbeck:
🎉  Looks like we've got all the tests passing now. @mamercad Let me know if there's anything else you need for this.

mamercad:
> 🎉 Looks like we've got all the tests passing now. @mamercad Let me know if there's anything else you need for this.

Should be good, thanks so very much! Looking forward to your continued help 😄

plugins/modules/digital_ocean_droplet_info.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+DOCUMENTATION = r'''
+---
+module: digital_ocean_droplet_info
+short_description: Gather information about DigitalOcean Droplets
+description:
+    - This module can be used to gather information about Droplets.
+author: "Tyler Auerbeck (@tylerauerbeck)"
+version_added: 1.2.0
```
mamercad, line 18: Forgot this, next version will be [1.4.0](https://github.com/ansible-collections/community.digitalocean/releases) if you wouldn't mind changing it -- or, I can fix it up in the `release` branch if I remember.

mamercad (changes requested):

plugins/modules/digital_ocean_droplet_info.py
mamercad: And, apologies for the back-and-forth, I've been multitasking too much lately.

tylerauerbeck: No problems! Fixed.

mamercad (approved):
Again, thanks for this!

Issue #13, opened 2020-08-05
Mitigate sanity test ignores by fixing the actual problems


geerlingguy:
##### SUMMARY

Currently the `sanity` tests for this module have a number of rules ignored (16 total) in https://github.com/ansible-collections/community.digitalocean/blob/main/tests/sanity/ignore-2.11.txt

It would be nice if we could get that down to zero—or just a couple.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

Tests

##### ANSIBLE VERSION

N/A

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

mamercad:
Agree, working on it.

gundalow:
Cleaned up by #54

Now we only have `ignore-2.9.txt` https://github.com/ansible-collections/community.digitalocean/tree/main/tests/sanity

PR #54, opened 2021-04-10
Fix sanity tests instead of ignoring them

mamercad:
##### SUMMARY

Stop ignoring sanity tests and fix the problems #13 .

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

* digital_ocean
* digital_ocean_droplet
* digital_ocean_floating_ip
* digital_ocean_sshkey


mamercad:
Bleh, why is the behavior different locally, sigh.

tests/sanity/ignore-2.9.txt
```diff
@@ -1,16 +0,0 @@
-plugins/modules/digital_ocean.py validate-modules:deprecation-mismatch
```
felixfontein, line 1: The `deprecation-mismatch` lines in `ignore-2.9.txt` need to stay.

changelogs/fragments/13-fix-sanity-tests.yaml
```diff
@@ -0,0 +1,5 @@
+bugfixes:
+  - digital_ocean - fixing sanity tests (https://github.com/ansible-collections/community.digitalocean/issues/13).
```
felixfontein, line 2: Changing documentation is not really noteworthy. What you should document in the changelog are user-facing changes, i.e. any modification you made to `argument_spec`.

tests/sanity/ignore-2.9.txt
felixfontein: (`invalid-documentation` and `missing-main-call` for the deprecated modules as well, probably.)

plugins/modules/digital_ocean.py
```diff
@@ -433,7 +448,7 @@ def main():
             size_id=dict(),
             image_id=dict(),
             region_id=dict(),
-            ssh_key_ids=dict(type='list', no_log=False),
+            ssh_key_ids=dict(type='list', elements='str', no_log=False),
```
felixfontein, line 451: This change potentially should have a changelog fragment ("``ssh_key_ids`` list entries are now validated to be strings").

plugins/modules/digital_ocean_droplet.py
```diff
@@ -315,22 +330,23 @@ def main():
             oauth_token=dict(
                 aliases=['API_TOKEN'],
                 no_log=True,
-                fallback=(env_fallback, ['DO_API_TOKEN', 'DO_API_KEY', 'DO_OAUTH_TOKEN'])
+                fallback=(env_fallback, ['DO_API_TOKEN', 'DO_API_KEY', 'DO_OAUTH_TOKEN']),
+                required=True,
```
felixfontein, line 334: This change definitely needs a changelog fragment, it's a potentially breaking change (except in case the module didn't work before when it wasn't specified, then it's a bugfix).

plugins/modules/digital_ocean_droplet.py
```diff
@@ -315,22 +330,23 @@ def main():
...
             ),
             name=dict(type='str'),
             size=dict(aliases=['size_id']),
             image=dict(aliases=['image_id']),
             region=dict(aliases=['region_id']),
-            ssh_keys=dict(type='list', no_log=False),
+            ssh_keys=dict(type='list', elements='str', no_log=False),
```
felixfontein, line 340: This and the other `elements` additions should also have a changelog fragment.

plugins/modules/digital_ocean_droplet.py
mamercad: Looks like we don't need to require it since the module fails gracefully if it's undefined:

```
❯ env | grep -E 'DO_|API_'
❯ ansible-playbook -i localhost, -c local droplet.yml -v 2>/dev/null | grep msg:
  msg: Failed to login using API token, please verify validity of API token.
```

mamercad: Looks like we don't need to require it since the module fails gracefully if it's undefined:

```
❯ env | grep -E 'DO_|API_'
❯ ansible-playbook -i localhost, -c local droplet.yml -v 2>/dev/null | grep msg:
  msg: Failed to login using API token, please verify validity of API token.
```

plugins/modules/digital_ocean_droplet.py
mamercad: Cool, I think I have these all squared away now.

plugins/modules/digital_ocean_droplet.py
felixfontein: In that case it's not a breaking change :) Requiring it is still a good idea, so the error handling is more uniform.

felixfontein (approved):

mamercad:
As always, thanks so much for the help along the way @felixfontein, you have such a great eye.

PR #57, opened 2021-04-11
Return zone records when creating domain

### LINKED ISSUE #46: Zone file not returned on initial domain creation
Sorry for taking so long on this. Unfortunately, it seems like "something else is actually broken".

Take a look at this output: https://gist.github.com/mamercad/8effa85bd0381ab164682b78ca5b6a88

Here's the playbook: https://gist.github.com/mamercad/afe7db39085aa88b4833224a486f4b4b

It seems like there's a different problem: the first run shows "changed" (which is correct since they domain didn't exist), but, it doesn't give back the zone information (this seems like the *actual* bug). The subsequent run, without changing the play, show as "ok" and return the zone information (which is correct since nothing is changing).

Let's look into this a bit closer.

_Originally posted by @mamercad in https://github.com/ansible-collections/community.digitalocean/issues/22#issuecomment-807470228_

mamercad:
##### SUMMARY

Return zone records even if creating a new domain, fixes #46 .

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

* digital_ocean_domain

##### ADDITIONAL INFORMATION

Here's the behavior in this branch:

```paste below
❯ doctl compute domain delete digitalocean.letsbuildthe.cloud -f

<deleted the domain>

❯ ansible-playbook -i localhost, -c local dns.yml -v
PLAY [localhost] **********************************************************************************************************************************************************

TASK [Create a domain] ****************************************************************************************************************************************************
Sunday 11 April 2021  10:23:22 -0400 (0:00:00.053)       0:00:00.053 **********
changed: [localhost] => changed=true
  ansible_facts:
    discovered_interpreter_python: /usr/bin/python3
  domain:
    domain:
      name: digitalocean.letsbuildthe.cloud
      ttl: 1800
      zone_file: |-
        $ORIGIN digitalocean.letsbuildthe.cloud.
        $TTL 1800
        digitalocean.letsbuildthe.cloud. IN SOA ns1.digitalocean.com. hostmaster.digitalocean.letsbuildthe.cloud. 1618151005 10800 3600 604800 1800
        digitalocean.letsbuildthe.cloud. 1800 IN NS ns1.digitalocean.com.
        digitalocean.letsbuildthe.cloud. 1800 IN NS ns2.digitalocean.com.
        digitalocean.letsbuildthe.cloud. 1800 IN NS ns3.digitalocean.com.


❯ ansible-playbook -i localhost, -c local dns.yml -v
PLAY [localhost] **********************************************************************************************************************************************************

TASK [Create a domain] ****************************************************************************************************************************************************
Sunday 11 April 2021  10:23:27 -0400 (0:00:00.052)       0:00:00.052 **********
ok: [localhost] => changed=false
  ansible_facts:
    discovered_interpreter_python: /usr/bin/python3
  domain:
    domain:
      name: digitalocean.letsbuildthe.cloud
      ttl: 1800
      zone_file: |-
        $ORIGIN digitalocean.letsbuildthe.cloud.
        $TTL 1800
        digitalocean.letsbuildthe.cloud. IN SOA ns1.digitalocean.com. hostmaster.digitalocean.letsbuildthe.cloud. 1618151005 10800 3600 604800 1800
        digitalocean.letsbuildthe.cloud. 1800 IN NS ns1.digitalocean.com.
        digitalocean.letsbuildthe.cloud. 1800 IN NS ns2.digitalocean.com.
        digitalocean.letsbuildthe.cloud. 1800 IN NS ns3.digitalocean.com.

PLAY RECAP ****************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

Here's the behavior out of `main` right now:

```
❯ ansible-playbook -i localhost, -c local dns.yml -v
Using /home/mark/.ansible.cfg as config file
[WARNING]: running playbook inside collection community.digitalocean

PLAY [localhost] **********************************************************************************************************************************************************

TASK [Create a domain] ****************************************************************************************************************************************************
Sunday 11 April 2021  10:28:48 -0400 (0:00:00.048)       0:00:00.048 **********
changed: [localhost] => changed=true
  ansible_facts:
    discovered_interpreter_python: /usr/bin/python3
  domain:
    name: digitalocean.letsbuildthe.cloud
    ttl: null
    zone_file: null

PLAY RECAP ****************************************************************************************************************************************************************
localhost                  : ok=1    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0

Playbook run took 0 days, 0 hours, 0 minutes, 2 seconds
Sunday 11 April 2021  10:28:50 -0400 (0:00:01.987)       0:00:02.036 **********
===============================================================================
Create a domain ---------------------------------------------------------------------------------------------------------------------------------------------------- 1.99s
```

### DIFF
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -171,7 +171,7 @@ def run(module):
                 # Arguably, it's nice to see the records versus null, so, we'll just try a
                 # few times before giving up and returning null.
                 for i in range(5):
-                    if do_manager.domain_record()['domain']['zone_file'] != 'null':
+                    if do_manager.domain_record()['domain']['zone_file'] is not None:
                         module.exit_json(changed=True, domain=do_manager.domain_record()['domain'])
                     time.sleep(3)
                 module.exit_json(changed=True, domain=domain)
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -75,7 +75,7 @@ EXAMPLES = r'''
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.community.digitalocean.plugins.module_utils.digital_ocean import DigitalOceanHelper
-from time import sleep
+import time
 
 
 class DoManager(DigitalOceanHelper, object):
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -77,6 +77,8 @@ from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.community.digitalocean.plugins.module_utils.digital_ocean import DigitalOceanHelper
 import time
 
+ZONE_FILE_ATTEMPTS = 5
+ZONE_FILE_SLEEP    = 3
 
 class DoManager(DigitalOceanHelper, object):
     def __init__(self, module):
@@ -170,10 +172,10 @@ def run(module):
                 #
                 # Arguably, it's nice to see the records versus null, so, we'll just try a
                 # few times before giving up and returning null.
-                for i in range(5):
+                for i in range(ZONE_FILE_ATTEMPTS):
                     if do_manager.domain_record()['domain']['zone_file'] is not None:
                         module.exit_json(changed=True, domain=do_manager.domain_record()['domain'])
-                    time.sleep(3)
+                    time.sleep(ZONE_FILE_SLEEP)
                 module.exit_json(changed=True, domain=domain)
         else:
             records = do_manager.all_domain_records()
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -162,7 +162,8 @@ def core(module):
             if 'message' in domain:
                 module.fail_json(changed=False, msg=domain['message'])
             else:
-                module.exit_json(changed=True, domain=domain)
+                records = do_manager.all_domain_records()
+                module.exit_json(changed=True, domain=do_manager.domain_record())
         else:
             records = do_manager.all_domain_records()
             if module.params.get('ip'):
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -73,10 +73,9 @@ EXAMPLES = r'''
 
 '''
 
-import traceback
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.community.digitalocean.plugins.module_utils.digital_ocean import DigitalOceanHelper
-from ansible.module_utils._text import to_native
+from time import sleep
 
 
 class DoManager(DigitalOceanHelper, object):
@@ -151,7 +150,7 @@ class DoManager(DigitalOceanHelper, object):
         return json['domain_record']
 
 
-def core(module):
+def run(module):
     do_manager = DoManager(module)
     state = module.params.get('state')
 
@@ -162,8 +161,20 @@ def core(module):
             if 'message' in domain:
                 module.fail_json(changed=False, msg=domain['message'])
             else:
-                records = do_manager.all_domain_records()
-                module.exit_json(changed=True, domain=do_manager.domain_record())
+                # We're at the mercy of a backend process which we have no visibility into:
+                # https://developers.digitalocean.com/documentation/v2/#create-a-new-domain
+                #
+                # In particular: "Keep in mind that, upon creation, the zone_file field will
+                # have a value of null until a zone file is generated and propagated through
+                # an automatic process on the DigitalOcean servers."
+                #
+                # Arguably, it's nice to see the records versus null, so, we'll just try a
+                # few times before giving up and returning null.
+                for i in range(5):
+                    if do_manager.domain_record()['domain']['zone_file'] != 'null':
+                        module.exit_json(changed=True, domain=do_manager.domain_record()['domain'])
+                    time.sleep(3)
+                module.exit_json(changed=True, domain=domain)
         else:
             records = do_manager.all_domain_records()
             if module.params.get('ip'):
@@ -209,10 +220,7 @@ def main():
         ),
     )
 
-    try:
-        core(module)
-    except Exception as e:
-        module.fail_json(msg=to_native(e), exception=traceback.format_exc())
+    run(module)
 
 
 if __name__ == '__main__':
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -80,6 +80,7 @@ import time
 ZONE_FILE_ATTEMPTS = 5
 ZONE_FILE_SLEEP = 3
 
+
 class DoManager(DigitalOceanHelper, object):
     def __init__(self, module):
         super(DoManager, self).__init__(module)
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -78,7 +78,7 @@ from ansible_collections.community.digitalocean.plugins.module_utils.digital_oce
 import time
 
 ZONE_FILE_ATTEMPTS = 5
-ZONE_FILE_SLEEP    = 3
+ZONE_FILE_SLEEP = 3
 
 class DoManager(DigitalOceanHelper, object):
     def __init__(self, module):
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -173,8 +173,11 @@ def run(module):
                 # Arguably, it's nice to see the records versus null, so, we'll just try a
                 # few times before giving up and returning null.
                 for i in range(ZONE_FILE_ATTEMPTS):
-                    if do_manager.domain_record()['domain']['zone_file'] is not None:
-                        module.exit_json(changed=True, domain=do_manager.domain_record()['domain'])
+                    record = do_manager.domain_record()
+                    if record is not None and 'domain' in record:
+                        domain = record['domain']
+                        if domain is not None and 'zone_file' in domain:
+                            module.exit_json(changed=True, domain=domain)
                     time.sleep(ZONE_FILE_SLEEP)
                 module.exit_json(changed=True, domain=domain)
         else:
--- a/changelogs/fragments/46-fix-domain-create-return-records.yaml
+++ b/changelogs/fragments/46-fix-domain-create-return-records.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - digital_ocean_domain - return zone records when creating a new zone (https://github.com/ansible-collections/community.digitalocean/issues/46).


mamercad:
In case it's needed, here's the playbook I'm using:

```
❯ cat dns.yml
---
- hosts: localhost
  connection: local
  gather_facts: no
  become: no

  tasks:

    - name: Create a domain
      community.digitalocean.digital_ocean_domain:
        state: present
        name: digitalocean.letsbuildthe.cloud
```

plugins/modules/digital_ocean_domain.py
```diff
@@ -162,7 +162,8 @@ def core(module):
             if 'message' in domain:
                 module.fail_json(changed=False, msg=domain['message'])
             else:
-                module.exit_json(changed=True, domain=domain)
+                records = do_manager.all_domain_records()
+                module.exit_json(changed=True, domain=do_manager.domain_record())
```
tadeboro, line 166: The [official docs](https://developers.digitalocean.com/documentation/v2/#create-a-new-domain) say this at the end:

> Keep in mind that, upon creation, the zone_file field will have a value of null until a zone file is generated and propagated through an automatic process on the DigitalOcean servers.

So this code just happens to work because listing all domain records and fetching the selected record once again gives the backend time to propagate the changes. Unfortunately, this method is as reliable as resolving race conditions by adding `sleep 3` to bash scripts. This will fail again if the DO backend becomes busy and takes more than a few seconds to update the zone file.

If we want to solve this reliably, we need to fetch data from API in a look until the zone file is present (with some upper bound of retries to avoid looping forever, of course).

tadeboro (changes requested):
I am afraid the current implementation will not provide a reliable fix to the issue at hand.

plugins/modules/digital_ocean_domain.py
mamercad: Ah, good find, I'll see what I can figure out.

plugins/modules/digital_ocean_domain.py
```diff
@@ -162,6 +161,19 @@ def core(module):
...
+                # https://developers.digitalocean.com/documentation/v2/#create-a-new-domain
+                #
+                # In particular: "Keep in mind that, upon creation, the zone_file field will
+                # have a value of null until a zone file is generated and propagated through
+                # an automatic process on the DigitalOcean servers."
+                #
+                # Arguably, it's nice to see the records versus null, so, we'll just try a
+                # few times before giving up and returning null.
+                for i in range(5):
+                    if do_manager.domain_record()['domain']['zone_file'] != 'null':
```
tadeboro, line 174: The API returns back a JSON `null`, which maps to `None` after the parsing is over.

```suggestion
                    if do_manager.domain_record()['domain']['zone_file'] is not None:
```

tadeboro (changes requested):
I found just one little error in the code. Apart from that, I think this is ready to ship.

mamercad:
Thank you for helping with this @tadeboro !

plugins/modules/digital_ocean_domain.py
```diff
@@ -162,6 +161,19 @@ def core(module):
...
+                # https://developers.digitalocean.com/documentation/v2/#create-a-new-domain
+                #
+                # In particular: "Keep in mind that, upon creation, the zone_file field will
+                # have a value of null until a zone file is generated and propagated through
+                # an automatic process on the DigitalOcean servers."
+                #
+                # Arguably, it's nice to see the records versus null, so, we'll just try a
+                # few times before giving up and returning null.
+                for i in range(5):
+                    if do_manager.domain_record()['domain']['zone_file'] is not None:
```
mpontillo, line 174: I'm not sure it's safe to index the domain record like this.

The `domain_record()` function just returns the response JSON without checking for errors. Isn't it possible that `domain` or `zone_file` won't exist?

plugins/modules/digital_ocean_domain.py
```diff
@@ -162,6 +161,19 @@ def core(module):
...
+                        module.exit_json(changed=True, domain=do_manager.domain_record()['domain'])
```
mpontillo, line 175: Since it looks like the `domain_record()` function calls into the REST API, I think it might be better to use the object that was returned from the previous call. Maybe something more like:
```
record = do_manager.domain_record()
if 'domain' in record:
    domain = record['domain']
    if 'zone_file' in domain:
        module.exit_json(changed=True, domain=domain)

File: plugins/modules/digital_ocean_domain.py [hunk 5]
```diff
@@ -162,6 +161,19 @@ def core(module):
...
+                # We're at the mercy of a backend process which we have no visibility into:
+                # https://developers.digitalocean.com/documentation/v2/#create-a-new-domain
+                #
+                # In particular: "Keep in mind that, upon creation, the zone_file field will
+                # have a value of null until a zone file is generated and propagated through
+                # an automatic process on the DigitalOcean servers."
+                #
+                # Arguably, it's nice to see the records versus null, so, we'll just try a
+                # few times before giving up and returning null.
+                for i in range(5):
```
mpontillo, line 173: Consider making the number of retries a parameter (or at least a constant).

plugins/modules/digital_ocean_domain.py
```diff
@@ -162,6 +161,19 @@ def core(module):
...
+                    if do_manager.domain_record()['domain']['zone_file'] is not None:
+                        module.exit_json(changed=True, domain=do_manager.domain_record()['domain'])
+                    time.sleep(3)
```
mpontillo, line 176: Consider making the sleep time a parameter (or at least a constant).

plugins/modules/digital_ocean_domain.py
tadeboro: `if 'zone_file' in domain`  will be always true according to official API docs, so this implementation will not work as expected.

plugins/modules/digital_ocean_domain.py
tadeboro: The `domain` key can be missing if the response code is not 200, but I am not sure if fixing this is in the scope of this PR. I would say that we should fix this (and other similar issues in the remaining data fetching methods) in a separate follow-up PR.

plugins/modules/digital_ocean_domain.py
mpontillo: Oh, one more thing. Due to the lack of error checking from the `domain_record()` function, you should probably also and a checks for None:
```
record = do_manager.domain_record()\
if record is not None and 'domain' in record:
    domain = record['domain']
    if domain is not None and 'zone_file' in domain:
        module.exit_json(changed=True, domain=domain)
```

mamercad: I agree with both of you on this -- I'll add the existential checking for now. I like @tadeboro suggestion of revisiting this down the road and making the API calls more robust.

mpontillo: This really comes down to the question of "how do we know whether or not the async process has updated the record"? In the case of a "new domain", it's not that difficult: it's either present, or not. (I didn't check how that will look in the returned JSON; I was only worried about Python exploding when trying to dereference `None`.) In the case of an update, I suppose you'd have to check whether your updates are present or not.

tadeboro: Checking for an update might prove itself to be unfeasible here. It would probably mean parsing the contents of the zone_file context, which sounds like a big no ;)

plugins/modules/digital_ocean_domain.py
mpontillo: I know this section of code is out of date, so this comment may be irrelevant now, but I'll have to disagree here. This is new code that has the potential to cause a traceback in the Python code as-is. For example, if the return is `None` or an empty dictionary you might see a `KeyError` or `TypeError`.

<details name="blah">
    <summary>Example REPL</summary>

```
>>> foo = {}
>>> foo['domain']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'domain'

>>> foo = None
>>> foo['domain']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not subscriptable
```

</details>

mpontillo: (that said, I'm fine with resolving this in a separate PR as long as it's not in a release.)

plugins/modules/digital_ocean_domain.py
```diff
@@ -208,10 +226,7 @@ def main():
         ),
     )
 
-    try:
-        core(module)
-    except Exception as e:
-        module.fail_json(msg=to_native(e), exception=traceback.format_exc())
+    run(module)
```
mpontillo, line 229: I'm curious about this change; catching `Exception` is certainly overly broad and should be avoided, but without it, how is it handled? Does the framework transparently handle the exception and similarly issue a `fail_json` or similar?

mpontillo (approved):
I think this is safe to merge now; thanks for the fixes!

I guess there are still some unresolved issues around how to know if the domain was really updated or not, but we can iterate on those.

plugins/modules/digital_ocean_domain.py
mpontillo: I see the code was changed as I suggested below, so there is no more traceback/regression potential here.

Sorry for the added noise on this PR. I'll resolve this conversation.

plugins/modules/digital_ocean_domain.py
mamercad: Seems like it ... I forced an `Exception` and got this:

```
TASK [Create a domain] **********************************************************************************************************************************
Tuesday 13 April 2021  18:25:39 -0400 (0:00:00.068)       0:00:00.068 ********* 
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: Exception: foo
[WARNING]: Platform darwin on host localhost is using the discovered Python interpreter at /usr/bin/python, but future installation of another Python
interpreter could change the meaning of that path. See https://docs.ansible.com/ansible/2.10/reference_appendices/interpreter_discovery.html for more
information.
fatal: [localhost]: FAILED! => changed=false 
  ansible_facts:
    discovered_interpreter_python: /usr/bin/python
  module_stderr: |-
    Traceback (most recent call last):
      File "/Users/mark/.ansible/tmp/ansible-tmp-1618352739.748239-91987-140231147269929/AnsiballZ_digital_ocean_domain.py", line 102, in <module>
        _ansiballz_main()
      File "/Users/mark/.ansible/tmp/ansible-tmp ...
```

...it's neatly handling it as a module failure and dumping the traceback to `module_stderr`.

tadeboro: Well, things are really simple. Ansible packages the module into a standalone application and runs it. If everything goes as planned, the application exits with a status 0 and prints serialized results to stdout.

If things go haywire (like in the case of non-handled exception), the program exits with a non-zero status and Ansible simply captures the stdout and stderr of the process.

So, the rule of thumb our team has: only catch errors that we expect and can provide a better diagnostic message to the user. If we can only return stacktrace, we have a bug in our code and the module should fail in an ugly way so we can fix it. For example, https://github.com/ansible-collections/servicenow.itsm/blob/main/docs/module_template.md is how we usually structure our code.

mamercad: > Well, things are really simple. Ansible packages the module into a standalone application and runs it. If everything goes as planned, the application exits with a status 0 and prints serialized results to stdout.
> 
> If things go haywire (like in the case of non-handled exception), the program exits with a non-zero status and Ansible simply captures the stdout and stderr of the process.
> 
> So, the rule of thumb our team has: only catch errors that we expect and can provide a better diagnostic message to the user. If we can only return stacktrace, we have a bug in our code and the module should fail in an ugly way so we can fix it. For example, https://github.com/ansible-collections/servicenow.itsm/blob/main/docs/module_template.md is how we usually structure our code.

Thanks for sharing this @tadeboro, I ⭐ 'd the template, thank you. Over time, I'd like to scour this codebase with an emphasis on consistency. I think we're going to lots of refactoring for 2.0 -- I've not liked `digital_ocean` (versus `digitalocean` for quite some time). Also, I don't really like the redundancy in the module names, e.g., I'd much rather see `community.digitalocean.droplet` versus `community.digitalocean.digital_ocean_droplet`.

PR #67, opened 2021-04-20
Enable Unit tests

gundalow:
##### SUMMARY
* Enable Unit tests (with code coverage)
* Tidy up integration tests so only invoke ansible-test once (avoids missing test, and speeds up)

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/.github/workflows/ansible-test-integration.yml
+++ b/.github/workflows/ansible-test-integration.yml
@@ -3,15 +3,15 @@ on:
   # ! Don't enable pwn requests !
   # pull_request:
   push:
-    paths:
-      - 'plugins/**'
-      - 'tests/**'
-      - '.github/workflows/ansible-test-*.yml'
     branches:
-      - ci
+      - ci*
       - main
   schedule:
     - cron: '10 6 * * *'
+
+env:
+  NAMESPACE: community
+  COLLECTION_NAME: digitalocean
 
 jobs:
 
@@ -21,13 +21,13 @@ jobs:
 
     defaults:
       run:
-        working-directory: ansible_collections/community/digitalocean
+        working-directory: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
 
     steps:
       - name: Clone the repo
         uses: actions/checkout@v2
         with:
-          path: ansible_collections/community/digitalocean
+          path: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
 
       - name: Set up Python 3.8
         uses: actions/setup-python@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/ansible-test-sanity.yml
```diff
@@ -1,17 +1,12 @@
...
-      - 'tests/**'
-      - '.github/workflows/ansible-test-*.yml'
   push:
-    paths:
-      - 'plugins/**'
-      - 'tests/**'
-      - '.github/workflows/ansible-test-*.yml'
   schedule:
     - cron: '0 6 * * *'
+env:
```
mamercad, line 7: Nit: Empty line before `env`?

.github/workflows/ansible-test-unit.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      # OPTIONAL If your unit test requires Python libraries from other collections
+      # Install them like this
+      - name: Install collection dependencies
+        run: ansible-galaxy collection install ansible.netcommon ansible.utils -p .
+
+      # Run the unit tests
+      - name: Run unit test
+        run: ansible-test units -v --color --docker --coverage
+        working-directory: ./ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
```
mamercad, line 50: Nit: Inconsistent use of leading `./` between lines 50 and 55.

.github/workflows/ansible-test-unit.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      # ansible-test support producing code coverage date
+      - name: Generate coverage report
+        run: ansible-test coverage xml -v --requirements --group-by command --group-by version
+        working-directory: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
+
+      # See the reports at https://codecov.io/gh/GITHUBORG/REPONAME
```
mamercad, line 57: How about replacing the comment with the actual?

```
# See the reports at https://app.codecov.io/gh/ansible-collections/community.digitalocean
```

.github/workflows/ansible-test-integration.yml
```diff
@@ -3,16 +3,16 @@ on:
   # ! Don't enable pwn requests !
   # pull_request:
   push:
-    paths:
```
gundalow, line 6: Always run the tests, otherwise codecoverage results can't be compared

.github/workflows/ansible-test-integration.yml
```diff
@@ -3,16 +3,16 @@ on:
   # ! Don't enable pwn requests !
   # pull_request:
```
mamercad, line 4: This is on me, perhaps we should say a bit more about what a "pwn request" is -- i.e., "We don't want to turn this on because a malicious PR could exfiltrate secrets by exploiting the Workflows."

.github/workflows/ansible-test-integration.yml
```diff
@@ -3,16 +3,16 @@ on:
...
-      - 'tests/**'
-      - '.github/workflows/ansible-test-*.yml'
     branches:
-      - ci
+      - ci*
       - main
   schedule:
     - cron: '10 6 * * *'
 
+env:
```
gundalow, line 12: Move closer to https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml so it's easier to update in the future

.github/workflows/ansible-test-integration.yml
```diff
@@ -45,92 +45,13 @@ jobs:
...
+        run: ansible-test integration --docker -v --color --retry-on-error --continue-on-error --python 3.8 --diff --coverage
 
       # ansible-test support producing code coverage date
       - name: Generate coverage report
-        run: |
-          mkdir -p tests/output/coverage
-          ansible-test coverage xml -v --requirements --group-by command --group-by version
-        working-directory: ansible_collections/community/digitalocean
+        run: ansible-test coverage xml -v --requirements --group-by command --group-by version
+        working-directory: ./ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
```
mamercad, line 54: Curious why we have to set the `working-directory` here when it's set in `defaults.run.working-directory`?

mamercad (approved):
I believe all I have are suggestions, questions and nitpicks -- I won't block on that.

.github/workflows/ansible-test-integration.yml
gundalow: Someone could still raise a PR and add that line back in so tests would be run.
Though the integration tests would fail as DO_API secret isn't exposed to forks.

.github/workflows/ansible-test-integration.yml
gundalow: Ah, hadn't noticed you'd set this. Should likely roll that into https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml

PR #97, opened 2021-05-23
Actually activate inventory caching

tadeboro:
##### SUMMARY

The inventory plugin advertises support for caching, but the implementation was missing the required pieces to make it work properly.

This commit split the inventory construction into two stages. The first stage is responsible for fetching remote data (interacting with the DigitalOcean's API). The second stage extracts the relevant data from the API responses and populated the inventory.

Caching support then came almost for free. All we had to do is only fetch data if:

 1. User does not want to use cache at all.
 2. User does use cache and wants data refreshed.
 3. User does use cache, but the cache is empty.

Some additional logic makes sure we write the data to cache if data refresh is requested, but nothing too complex.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`community.digitalocean.digitalocean` inventory plugin

### DIFF
--- a/changelogs/fragments/97-wire-up-inventory-cache.yml
+++ b/changelogs/fragments/97-wire-up-inventory-cache.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - digitalocean inventory plugin - wire up advertised caching functionality
+    (https://github.com/ansible-collections/community.digitalocean/pull/97).
--- a/plugins/inventory/digitalocean.py
+++ b/plugins/inventory/digitalocean.py
@@ -170,12 +170,12 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         return payload
 
-    def _populate(self):
+    def _populate(self, records):
         attributes = self.get_option('attributes')
         var_prefix = self.get_option('var_prefix')
         strict = self.get_option('strict')
         host_filters = self.get_option('filters')
-        for record in self._get_payload():
+        for record in records:
 
             host_name = record.get('name')
             if not host_name:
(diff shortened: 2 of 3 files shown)


tadeboro:
Not sure what codecov did here, but those numbers are definitely broken. It looks like we are comparing full coverage with partial coverage (only lines that I changed), but since I do not use codecov personally, I have no experience with these things.

mamercad:
@gundalow Can you help us with the failing `codecov/patch` test?

mamercad (approved):
Thanks for fixing this, LGTM:

```bash
❯ grep cache digitalocean.yml
cache: yes
cache_plugin: jsonfile
cache_timeout: 300
cache_connection: /tmp/do_inventory
cache_prefix: do_

❯ head /tmp/do_inventory/do_community.digitalocean.digitalocean_bccfds_7fdd2
[
    {
        "backup_ids": [],
        "created_at": "2021-05-24T23:40:20Z",
        "disk": 25,
        "features": [
            "private_networking"
        ],
        "id": 247480313,
        "image": {
```

mamercad:
@tadeboro Want to add a quick changelog fragment?

mamercad (changes requested):
Changelog fragment please.

tadeboro:
@mamercad I wanted to add a changelog fragment once I knew the PR number, but then the code coverage distracted me ;)

gundalow:
> @gundalow Can you help us with the failing `codecov/patch` test?

Could you try changing `codecov.yml` to the following:

```yaml
coverage:
  status:
    project:
      default:
        informational: true # Don't fail CI if percentage drops
    patch:
      default:
        informational: true # Don't fail CI if percentage drops

# Don't count lines of code in these directories
ignore:
    - tests
fixes:
  - "ansible_collections/community/digitalocean/::"
```

tadeboro:
I rebased the changes onto the ` main` branch. The codecov is still complaining because the updated lines are not covered by unit tests. And while I can add tests to cover them just to make codecov happy, I would rather not. Why? Because I will need to mock 90% of the inventory plugin functionality just to cover three if statements. And such tests (tests that do not actually test any surrounding code) are worthless in the long run since they do not adapt to the changes in the core.

mamercad (approved):
Thanks for fixing this!

PR #98, opened 2021-05-24
Support templating of the API token

maxtruxa:
##### SUMMARY
This PR adds support for templates in the API token field. This mechanism allows retrieving the password from arbitrary sources (for example, a password manager).

##### TODO
- [x] Add documentation.
- [x] Add tests.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Inventory plugin

##### ADDITIONAL INFORMATION
Example script `get-do-token.sh` that just writes the API token to stdout:
```sh
#!/usr/bin/env sh

# You can do whatever you want here.
echo 'your-api-token-here'
```
Inventory config `inventory_digitalocean.yml`, referencing the script:
```yml
api_token: '{{ lookup("pipe", "./get-do-token.sh") }}'
```

### DIFF
--- a/changelogs/fragments/98-api-token-command.yml
+++ b/changelogs/fragments/98-api-token-command.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - digitalocean - support templates in API tokens when using the dynamic inventory plugin (https://github.com/ansible-collections/community.digitalocean/pull/98).
--- a/plugins/inventory/digitalocean.py
+++ b/plugins/inventory/digitalocean.py
@@ -31,6 +31,7 @@ options:
   api_token:
     description:
      - DigitalOcean OAuth token.
+     - Template expressions can be used in this field.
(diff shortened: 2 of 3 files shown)


mamercad:
This is a neat idea, thank's for contributing.

mamercad:
I wonder if there's a function we can call to trigger interpolation, like this:

```yaml
api_token: "{{ lookup('pipe', './get-do-token.sh') }}"
```

maxtruxa:
Oh that is actually a much better idea. That shouldn't be too hard to implement, since it's being used all over the place in Ansible. I'll look into it later today.

maxtruxa:
@mamercad Done!

I changed the example to showcase the use of a `lookup`, because the old example was broken.

mamercad (approved):
This is great -- thanks so very much! 🎉

PR #102, opened 2021-05-27
Pass only prefixed attributes to filters

maxtruxa:
##### SUMMARY
This PR is mutually exclusive with #103 .

In contrast to the jinja expressions in `compose`, `groups` and `keyed_groups`, `filters` was given *all* droplet attributes without a prefix. My bad for not catching that.

Note that this will break anyone who started using this feature since the release of version 1.5.0. If that is a no-go, I can amend the PR to keep the previous inconsistent behavior as well.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Inventory plugin

##### ADDITIONAL INFORMATION
Before:
```yml
attributes:
  - tags
filters:
  - '"test" in tags'
```
or even without the attributes:
```yml
filters:
  - '"test" in tags'
```

After:
```yml
attributes:
  - tags
filters:
  - '"test" in do_tags'
```

### DIFF
--- a/changelogs/fragments/102-filters-prefixed-vars.yml
+++ b/changelogs/fragments/102-filters-prefixed-vars.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - digitalocean inventory plugin - attributes available to filters are limited
+    to explicitly required attributes and are prefixed with ``var_prefix``
+    (https://github.com/ansible-collections/community.digitalocean/pull/102).
--- a/plugins/inventory/digitalocean.py
+++ b/plugins/inventory/digitalocean.py
@@ -71,6 +71,7 @@ options:
     type: list
     elements: str
     default: []
+    version_added: '1.5.0'
 '''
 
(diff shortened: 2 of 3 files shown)


tadeboro (changes requested):
I think this PR is superior compared to #103 because it preserves the consistency of behavior. And although this is indeed an API break, I would say that the original behavior was flawed, and thus this is a fix that sets things straight. Plus, the Ansible PyPI package did not include the "broken" version of the plugin yet, so not many users were exposed to it.

But this PR does need a documentation fragment that describes the change, this is why I am requesting changes in my review.

maxtruxa:
@tadeboro Thanks for the feedback! I added the missing changelog fragment.

tadeboro (approved):

mamercad (approved):
Thank you very much for this, want to close #103?

maxtruxa:
Closed #103. Thank you!

PR #103, opened 2021-05-27
Expose unprefixed attributes to inventory config

maxtruxa:
##### SUMMARY
This PR is mutually exclusive with #102.

Going the complete opposite direction from #102, this allows the use of unprefixed attributes in all template fields.

The purpose of the `var_prefix` seems to be to avoid clashes with variables from other sources and make it more obvious where the variable came from. Both of these points don't apply in the context of the inventory plugin's config, since any host-/group-/role-/play-specific variables are loaded only after the inventory plugin is done processing.

With these changes you can use *all* droplet attributes without a prefix in `filters`, `compose`, `groups` and `keyed_groups`, no matter if you specified them in `attributes` or not. Prefixed versions of the attributes listed in `attributes` are still available (this is the old behavior).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Inventory plugin

##### ADDITIONAL INFORMATION
Before:
```yml
attributes:
  - id
  - name
  - tags
  - region
groups:
  development: '"development" in do_tags'
keyed_groups:
  - key: do_region.slug
    prefix: 'region'
    separator: '_'
leading_separator: false
compose:
  ansible_host: do_networks.v4 | selectattr('type', 'eq', 'public') | map(attribute='ip_address') | first
  distro: do_image.distribution | lower
filters:
  - '"test" in tags'
```
After (but note that the "before" version still works unchanged):
```yml
attributes:
  - id
  - name
  - tags
  - region
groups:
  development: '"development" in tags'
keyed_groups:
  - key: region.slug
    prefix: 'region'
    separator: '_'
leading_separator: false
compose:
  ansible_host: networks.v4 | selectattr('type', 'eq', 'public') | map(attribute='ip_address') | first
  distro: image.distribution | lower
filters:
  - '"test" in tags'
```

### DIFF
--- a/plugins/inventory/digitalocean.py
+++ b/plugins/inventory/digitalocean.py
@@ -71,6 +71,7 @@ options:
     type: list
     elements: str
     default: []
+    version_added: '1.5.0'
 '''
 
 EXAMPLES = r'''
@@ -113,6 +114,7 @@ from ansible.module_utils._text import to_native
 from ansible.module_utils.urls import Request
 from ansible.module_utils.six.moves.urllib.error import URLError, HTTPError
 from ansible.plugins.inventory import BaseInventoryPlugin, Constructable, Cacheable
+from ansible.utils.vars import combine_vars
 
 
 class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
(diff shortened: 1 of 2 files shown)


tadeboro (commented):
I think #102 is superior to this one.

PR #109, opened 2021-06-05
Fix info modules

### LINKED ISSUE #55: Info modules do not always return alist as documented
##### SUMMARY
Most of the info modules in the `community.digitalocean` Ansible Collection state that they return a list of resources, but the code revealed that some of those modules return a single resource in certain scenarios.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
info modules

##### LIST OF AFFECTED MODULES

  * https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/modules/digital_ocean_certificate_info.py
  * https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/modules/digital_ocean_domain_info.py
  * https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/modules/digital_ocean_firewall_info.py
  * https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/modules/digital_ocean_load_balancer_info.py
  * https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/modules/digital_ocean_snapshot_info.py
  * https://github.com/ansible-collections/community.digitalocean/blob/main/plugins/modules/digital_ocean_tag_info.py

tylerauerbeck:
##### SUMMARY
Fixes #55 . This ensures that info modules return a list of dicts regardless if searching for a single item or all items.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
`digital_ocean_certificate_info`
`digital_ocean_domain_info`
`digital_ocean_firewall_info`
`digital_ocean_load_balancer_info`
`digital_ocean_tag_info`

##### ADDITIONAL INFORMATION
This also adds tests for modules that didn't have them.


tests/integration/targets/digital_ocean_tag_info/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+      when:
+        - do_api_key is not defined
+        - do_api_key | length == 0
+
+    - name: Gather tag information
+      community.digitalocean.digital_ocean_tag_info:
+        oauth_token: "{{ do_api_key }}"
+      register: result
+
+    - name: Verify droplet info fetched
```
mamercad, line 16: Change `droplet` to `tag` and perhaps `info` to `information` (for consistency).

tests/integration/targets/digital_ocean_firewall_info/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+      when:
+        - do_api_key is not defined
+        - do_api_key | length == 0
+
+    - name: Gather firewall information
+      community.digitalocean.digital_ocean_firewall_info:
+        oauth_token: "{{ do_api_key }}"
+      register: result
+
+    - name: Verify droplet info fetched
```
mamercad, line 16: Change `droplet` to `firewall` and perhaps `info` to `information` (for consistency).

tests/integration/targets/digital_ocean_domain_info/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+      when:
+        - do_api_key is not defined
+        - do_api_key | length == 0
+
+    - name: Gather domain information
+      community.digitalocean.digital_ocean_domain_info:
+        oauth_token: "{{ do_api_key }}"
+      register: result
+
+    - name: Verify droplet info fetched
```
mamercad, line 16: Change `droplet` to `domain` and perhaps `info` to `information` (for consistency).

tests/integration/targets/digital_ocean_certificate_info/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+      when:
+        - do_api_key is not defined
+        - do_api_key | length == 0
+
+    - name: Gather certificate information
+      community.digitalocean.digital_ocean_certificate_info:
+        oauth_token: "{{ do_api_key }}"
+      register: result
+
+    - name: Verify droplet info fetched
```
mamercad, line 16: Change `droplet` to `certificate` and perhaps `info` to `information` (for consistency).

plugins/modules/digital_ocean_certificate_info.py
```diff
@@ -87,8 +88,7 @@ def core(module):
         if status_code != 200:
             module.fail_json(msg="Failed to retrieve certificates for DigitalOcean")
 
-        resp_json = response.json
-        certificate = resp_json['certificate']
+        certificate = [response.json['certificate']]
```
mamercad, line 91: Perhaps we should test keys before assuming they're there to mitigate potential exceptions.

mamercad (changes requested):
Few small things to fix up, thanks for cleaning up these modules!

plugins/modules/digital_ocean_certificate_info.py
tylerauerbeck: That makes sense. Do you want to do this as part of this PR or a follow up though? Because I think that is a fairly common thing that probably needs addressed across all of the other modules as well.

mamercad: > That makes sense. Do you want to do this as part of this PR or a follow up though? Because I think that is a fairly common thing that probably needs addressed across all of the other modules as well.

Yep, that's a good point, we'll be doing lots of changes in 2.x -- e.g., I'd like to get rid of the redundant `digital_ocean` prefix (in addition to being redundant, "DigitalOcean" is actually one word). So, we can leave it alone for now.

mamercad (approved):
Thanks, let's 🚢 this!

PR #111, opened 2021-06-13
New module: DigitalOcean Snapshot (Droplet and Volume)

### LINKED ISSUE #91: RFE: Snapshot Module
##### SUMMARY
Currently there is a module to get information about a snapshot, but not to take a snapshot itself. There currently isn't an individual endpoint (at least that I'm aware of) that allows you to create a snapshot. But this functionality does exist on both the block storage and droplet endpoints. It would nice to be able to control/interact with this just like the rest of the modules in this collection

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`digital_ocean_snapshot`

##### ADDITIONAL INFORMATION
N/A

mamercad:
##### SUMMARY

Fixes #91; create and delete Droplet and volume snapshots.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

* digital_ocean_snapshot

```yaml
- name: Snapshot a Droplet
  community.digitalocean.digital_ocean_snapshot:
    state: present
    snapshot_type: droplet
    snapshot_name: mysnapshot1
    droplet_id: 250329179
  register: result

- name: Delete a Droplet snapshot
  community.digitalocean.digital_ocean_snapshot:
    state: absent
    snapshot_type: droplet
    snapshot_id: 85905825
  register: result

- name: Snapshot a Volume
  community.digitalocean.digital_ocean_snapshot:
    state: present
    snapshot_type: volume
    snapshot_name: mysnapshot2
    volume_id: 9db5e329-cc68-11eb-b027-0a58ac144f91

- name: Delete a Volume snapshot
  community.digitalocean.digital_ocean_snapshot:
    state: absent
    snapshot_type: volume
    snapshot_id: a902cdba-cc68-11eb-a701-0a58ac145708
```

### DIFF
--- a/plugins/modules/digital_ocean_snapshot.py
+++ b/plugins/modules/digital_ocean_snapshot.py
@@ -80,15 +80,13 @@ EXAMPLES = r"""
     snapshot_type: droplet
     snapshot_name: mysnapshot1
     droplet_id: 250329179
-    oauth_token: "{{ lookup('ansible.builtin.env', 'DO_API_KEY') }}"
   register: result
 
 - name: Delete a Droplet snapshot
   community.digitalocean.digital_ocean_snapshot:
     state: absent
     snapshot_type: volume
-    snapshot_id: 250329179:
-    oauth_token: "{{ oauth_token }}"
+    snapshot_id: 85905825
   register: result
 
 - name: Snapshot a Volume
@@ -97,14 +95,12 @@ EXAMPLES = r"""
     snapshot_type: volume
     snapshot_name: mysnapshot2
     volume_id: 9db5e329-cc68-11eb-b027-0a58ac144f91
-    oauth_token: "{{ lookup('ansible.builtin.env', 'DO_API_KEY') }}"
 
 - name: Delete a Volume snapshot
   community.digitalocean.digital_ocean_snapshot:
     state: absent
     snapshot_type: volume
     snapshot_id: a902cdba-cc68-11eb-a701-0a58ac145708
-    oauth_token: "{{ lookup('ansible.builtin.env', 'DO_API_KEY') }}"
 """
 
 
@@ -116,7 +112,7 @@ result:
   sample:
     changed: true
     failed: false
-    msg: Created snapshot, snapshot 250329179
+    msg: Created snapshot, snapshot 85905825
 """
 
 
@@ -176,6 +172,7 @@ class DOSnapshot(object):
                 self.action_id = json["action"]["id"]
                 if self.wait:
                     json = self.wait_finished()
+                    raise Exception(str(json))
                     self.module.exit_json(
                         changed=True,
                         msg="Created snapshot, snapshot {0}".format(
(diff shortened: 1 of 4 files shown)


plugins/modules/digital_ocean_snapshot.py
```diff
@@ -0,0 +1,265 @@
...
+        droplet_id=dict(type="str"),
+        volume_id=dict(type="str"),
+        snapshot_id=dict(type="str"),
+        wait=dict(type="bool", default=True),
+        wait_timeout=dict(default=120, type="int"),
+    )
+    module = AnsibleModule(
+        argument_spec=argument_spec,
+        required_if=[
+            ("state", "present", ["snapshot_name"]),
```
tadeboro, line 254: Why is the name required here? As far as I can tell, the API does not require it and does not offer any means of managing snapshots by name.

plugins/modules/digital_ocean_snapshot.py
```diff
@@ -0,0 +1,265 @@
...
+            ("state", "absent", ["snapshot_id"]),
+            ("snapshot_type", "droplet", ["droplet_id"]),
+            ("snapshot_type", "volume", ["volume_id"]),
```
tadeboro, line 257: Are those two really needed? I would assume this could cause trouble when trying to delete snapshots because the user would need to provide a droplet ID. I would assume that dropped or volume IDs are only needed when the state is set to present.

tadeboro (commented):
I mostly reviewed the user-facing interface at this time and did not look at the implementation too carefully. I also noticed that the check mode is not supported. Was this done on purpose?

plugins/modules/digital_ocean_snapshot.py
```diff
@@ -0,0 +1,303 @@
...
+            if status != 200:
+                self.module.fail_json(
+                    msg="Unable to find action {0}, please file a bug".format(
+                        str(self.action_id)
+                    )
+                )
+            json = response.json
+            if json["action"]["status"] == "completed":
+                return json
+            time.sleep(min(10, end_time - time.monotonic()))
```
mpontillo, line 187: It is possible for `ValueError: sleep length must be non-negative` to happen here, if by the time this line is reached `time.monotonic()` is greater than `end_time`.

mpontillo (dismissed):
Looks generally good; one nit below on the usage of `time.sleep()`; this may be something that needs to be fixed in a lot of places throughout the code.

plugins/modules/digital_ocean_snapshot.py
mamercad: Ah yeah, the other idea is ... why are we trying to get savings here at all ... we're talking about a few seconds, heh.

mamercad:
> I mostly reviewed the user-facing interface at this time and did not look at the implementation too carefully. I also noticed that the check mode is not supported. Was this done on purpose?

Ah yes, check mode, I'll fix that.

mpontillo (approved):

PR #161, opened 2021-09-10
Release 1.10.0

mamercad:
##### SUMMARY

Release 1.10.0; new modules, minor changes, bugfixes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,32 @@ Community DigitalOcean Release Notes
 .. contents:: Topics
 
 
+v1.10.0
+=======
+
+Minor Changes
+-------------
+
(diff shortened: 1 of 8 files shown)


mamercad:
@felixfontein I had to manually fixup the changelog files after running `antsibull-changelog release`; the new modules `digital_ocean_cdn_endpoints_*` and `digital_ocean_monitoring_alerts_*` weren't picked up -- can you spot anything wrong?

mamercad:
> @mamercad I've tried this locally on my machine, and it worked. Is it possible that you had an outdated changelogs/.plugin-cache.yaml present which was created for 1.10.0 before these modules were added? If antsibull-changelog already has that file for the current release version, it will not re-create it automatically (but you can force it by specifying `--reload-plugins`).

Ah thank you, that must have been it -- rerunning with `--reload-plugins` did the trick!

mpontillo (approved):
🚢

PR #165, opened 2021-09-15
Serialize the integration tests

mamercad:
##### SUMMARY

It looks like we're probably running into race conditions with concurrent integration test runs. It's probably a better idea to serialize them, regardless.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

The integration tests.


gundalow (approved):
Nice, I didn't know about this GitHub Action feature.
Thank you for thinking about test stability.

gundalow:
Integration tests only run on `main`, so need to merge this PR, then check the GHA results for the default branch (not the PR)

PR #167, opened 2021-09-18
Improving integration testing (approval-based on PRs)

mamercad:
##### SUMMARY

By using `pull_request_target` and `environment` and `artifact`, I believe that this gives us a safe way (approval-based) to run integration tests on pull requests which doesn't expose secrets. I'd definitely like an audit of this, I'm by no means an expert in Github Workflows. I'd love to see integration tests running on pull requests, if possible (it'd make it much easier for reviewers to feel comfortable that things are working correctly). This will require setting up an "Environment" in this project and adding the API key secret to it.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Github Workflow `pull-request-integration` (new)

### DIFF
--- a/.github/workflows/pull-request-integration.yml
+++ b/.github/workflows/pull-request-integration.yml
@@ -1,0 +1,95 @@
+name: pull-request-integration
+
+on:
+  pull_request_target:
+    branches: [ main ]
+
+env:
+  NAMESPACE: community
+  COLLECTION_NAME: digitalocean
+
+concurrency:
+  group: cloud-integration-tests
+  cancel-in-progress: false
+
+jobs:
+
+  approve:
+    name: approve
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: approve
+        run: echo All pull requests need to be approved before running the integration tests.
+
+  secret:
+    name: secret
+    needs: [ approve ]
+    environment: integration
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: Write secret
+        shell: bash
+        run: echo "export DO_API_KEY='${{ secrets.DO_API_KEY }}'" >secret
+
+      - name: Upload secret
+        uses: actions/upload-artifact@v2
+        with:
+          name: secret
+          path: secret
+          retention-days: 0
+
+  integration:
+
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        working-directory: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
+
+    steps:
+
+      - name: Download secret
+        uses: actions/download-artifact@v2
+        with:
+          name: secret
+
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          path: ansible_collections/${{env.NAMESPACE}}/${{env.COLLECTION_NAME}}
+          ref: ${{ github.event.pull_request.head.sha }} # Check out the pull request
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install ansible-base (${{ matrix.ansible }})
+        run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
+
+      - name: Install ansible_collections.community.general
+        run: ansible-galaxy collection install community.general -p ../../
+
+      - name: Configure integration test run secrets
+        run: |
+          source ../../../secret
+          ./tests/utils/render.sh \
+            ./tests/integration/integration_config.yml.template \
+              > ./tests/integration/integration_config.yml
+
+      # Run the integration tests
+      - name: Run integration test
+        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --python ${{ matrix.python }} --docker --coverage
+
+        # ansible-test support producing code coverage date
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-integration.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Write secret
+        shell: bash
+        run: echo "export DO_API_KEY='${{ secrets.DO_API_KEY }}'" >secret
+
+      - name: Upload secret
+        uses: actions/upload-artifact@v2
+        with:
+          name: secret
+          path: secret
+          retention-days: 0
```
tadeboro, line 43: Why do we need those two jobs? I would assume that if the environment has the protections rules set, we just need to reference it in the integration job definition and GitHub will make sure nothing is executed unless someone approves the changes. (There is a high chance I am missing some information here because it has been a while since I read the GitHub docs for the last time.)

mamercad: Ah yeah, you're right, just using an `environment` should work here... I wasn't thinking (this test doesn't use `strategy` -- if it did, and the `environment` was set, it'd require approval for each combination). I was playing with this idea [here](https://github.com/ansible-collections/community.healthchecksio/blob/b6a9b24dc2790418ba8b6e6816bf5ce499cd189b/.github/workflows/pull-request-integration.yml#L44) and I think what I've actually figured out is a way to have a single approval while using `strategy`. I'll rework this PR to be simpler.

mamercad: Simplified it (pushed). Also, I set up an environment named `integration` with a secret named `DO_API_KEY` (not a real value yet), want to make sure what's there doesn't lend it itself to being [pwned](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

mamercad: I'm thinking we may still want two separate integration test workflows ... the "main" one runs on a schedule (and merges into main). We probably don't want to manually approve the daily runs.

tadeboro: Yes, the post-merge and scheduled runs can stay as they are now since we know they are safe to run. Only PRs should be treated with more care.

mamercad: Could you run this proposed change by any folks that you happen to know for their opinion on whether or not this is safe? The `ref: ${{ github.event.pull_request.head.sha }} # Check out the pull request` scares me, but, it's necessary so that we're running their commit (not `main`).

.github/workflows/pull-request-integration.yml
```diff
@@ -0,0 +1,67 @@
...
+          ref: ${{ github.event.pull_request.head.sha }} # Check out the pull request
+          persist-credentials: false
+
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Install ansible-base (${{ matrix.ansible }})
+        run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
```
tadeboro, line 43: Matrix is not defined anywhere, so this will probably fail to "compile".

.github/workflows/pull-request-integration.yml
```diff
@@ -0,0 +1,67 @@
...
+        env:
+          DO_API_KEY: ${{ secrets.DO_API_KEY }}
+        run: |
+          ./tests/utils/render.sh \
+            ./tests/integration/integration_config.yml.template \
+              > ./tests/integration/integration_config.yml
+
+      # Run the integration tests
+      - name: Run integration test
+        run: ansible-test integration -v --color --retry-on-error --continue-on-error --diff --python ${{ matrix.python }} --docker --coverage
```
tadeboro, line 58: Same thing here: the matrix of Python versions is not defined.

tadeboro (changes requested):
The config looks OK to me in general, but we will not know if things work as expected until we merge this and open a new PR.

.github/workflows/pull-request-integration.yml
mamercad: Ah yeah, bad copy pasta, I'll fix.

.github/workflows/pull-request-integration.yml
tadeboro: I did not talk about this with anyone yet (I encourage my team members to ignore me over the weekend ;)), but I think is safe to do because:

 1. Checking out the code at that specific commit does not alter the workflow definition (this is what `on: pull_request_target` gives us).
 2. Tests on that code are run only after someone approves the run and I would assume that the approver will make sure nothing malicious was done in the integration tests.

If you are not in a hurry to merge this, I can ping a few people tomorrow to have a look at the setup and give their feedback.

mamercad: Oh, hehe, I'm totally not in a hurry. I too feel that it's safe for these reasons, just looking for confirmation. We're on the same page, but, it doesn't hurt to get a few more eyes on it if possible -- as mentioned previously, I'm definitely not an expert in GitHub Workflows.

mamercad: Basically, it'd be sweet if PRs ran the integration tests so that we'd be more confident that things work correctly. I have a pretty large refactor of the the Droplet module open, and, before dropping that review on someone it'd be cool if they could see that the Droplet integration tests are still passing.

PR #182, opened 2021-10-27
Adding Project support

### LINKED ISSUE #171: RFE: Add project parameter to modules
SUMMARY
It would be nice to have the option of specifying the Project that the resource will be (created in/updated in/removed from) without changing the default project with the project module.

ISSUE TYPE
- Feature Idea

COMPONENT NAME
Module Project Parameter

ADDITIONAL INFORMATION
Instead of having to change the default project before creating, updating or removing a resource, it would be nice to have this as a module parameter. I can switch back and forth, so this is just a 'nice-to-have', but it makes for additional unnecessary tasks and increases likelihood of accidentally modifying the wrong project.

Ex:
  - name: Create Managed Postgres DB
    community.digitalocean.digital_ocean_database:
      oauth_token: "{{ do_api_token }}"
      state: present
      name: example-db
      engine: pg
      size: db-s-1vcpu-1gb
      region: tor1
      num_nodes: 1
      version: '12'
      **project: myproject1**

mamercad:
Fixes #171.

Adds Project assignment support (at the time of creation) for the following modules:

* digital_ocean_block_storage
* digital_ocean_database
* digital_ocean_domain
* digital_ocean_droplet
* digital_ocean_floating_ip
* digital_ocean_load_balancer

That is, the ability to associate these resources to [Projects](https://docs.digitalocean.com/products/projects/) when they're created.

Trying to make the smallest possible change, so, if the project isn't specified, the behavior will be unchanged (i.e., the resources will land in the "default" Project as before).

### DIFF
--- a/changelogs/fragments/171-add-project-support.yaml
+++ b/changelogs/fragments/171-add-project-support.yaml
@@ -1,0 +1,7 @@
+minor_changes:
+  - digital_ocean_block_storage - adding Project support (https://github.com/ansible-collections/community.digitalocean/issues/171).
+  - digital_ocean_database - adding Project support (https://github.com/ansible-collections/community.digitalocean/issues/171).
+  - digital_ocean_domain - adding Project support (https://github.com/ansible-collections/community.digitalocean/issues/171).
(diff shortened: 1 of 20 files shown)


danxg87 (approved):
LGTM

PR #227, opened 2022-02-27
AAAA record support for digital_ocean_domain

### LINKED ISSUE #226: AAAA record support for digital_ocean_domain
##### SUMMARY
The digital_ocean_domain module does not support creating a `AAAA` record which is an essential feature.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
digital_ocean_domain

mamercad:
##### SUMMARY

Fixes #226; I'm going to tackle individual DNS records with a new module, `digital_ocean_domain_record` (it'll be cleaner this way).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

* digital_ocean_domain

### DIFF
--- a/plugins/modules/digital_ocean_domain.py
+++ b/plugins/modules/digital_ocean_domain.py
@@ -41,7 +41,7 @@ options:
     description:
     - An 'AAAA' record for '@' ($ORIGIN) will be created with the value 'ip6'.  'ip6' is an IP version 6 address.
     type: str
-    aliases: ['ipv4', 'ipv6']
+    aliases: ['ipv6']
   project_name:
     aliases: ["project"]
     description:
@@ -308,7 +308,7 @@ def main():
         name=dict(type="str"),
         id=dict(aliases=["droplet_id"], type="int"),
         ip=dict(type="str", aliases=["ip4", "ipv4"]),
-        ip6=dict(type="str", aliases=["ip6", "ipv6"]),
+        ip6=dict(type="str", aliases=["ipv6"]),
         project_name=dict(type="str", aliases=["project"], required=False, default=""),
     )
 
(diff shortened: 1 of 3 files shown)


plugins/modules/digital_ocean_domain.py
```diff
@@ -155,7 +161,11 @@ def destroy_domain(self):
             return json
 
     def edit_domain_record(self, record):
-        params = {"name": "@", "data": self.module.params.get("ip")}
+        if self.module.params.get("ip"):
+            params = {"name": "@", "data": self.module.params.get("ip")}
+        if self.module.params.get("ip6"):
+            params = {"name": "@", "data": self.module.params.get("ip6")}
```
mpontillo, line 167: The `params` here is being overwritten; does this mean that this module only supports setting one type of apex record at a time?

mamercad: The [API docs](https://docs.digitalocean.com/reference/api/api-reference/#operation/create_domain_record) aren't overly clear, but yes, it looks like the `type` parameter can be *one of* `['A', 'AAAA', 'CNAME', ...]`.

mamercad: The [API docs](https://docs.digitalocean.com/reference/api/api-reference/#operation/create_domain_record) aren't overly clear, but yes, it looks like the `type` parameter can be *one of* `['A', 'AAAA', 'CNAME', ...]`.

mpontillo: So should we make this an `elif`, then, and document the idea that `ip` will be preferred over `ip6`? (as written, if both are supplied then only `ip6` will be updated)

mamercad: I don't think it matters; this module uses `mutually_exclusive=[("ip", "ip6")],` -- the module will fail if both arguments are specified.

mpontillo (approved):

PR #231, opened 2022-02-27
Adding digital_ocean_domain_record_info

### LINKED ISSUE #230: Module for Domain Records
##### SUMMARY

Create a module for [Domain Records](https://docs.digitalocean.com/reference/api/api-reference/#tag/Domain-Records).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

* digital_ocean_domain_record
* digital_ocean_domain_record_info

mamercad:
##### SUMMARY

Fixes #230.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

* digital_ocean_domain_record_info

### DIFF
--- a/plugins/modules/digital_ocean_domain_record_info.py
+++ b/plugins/modules/digital_ocean_domain_record_info.py
@@ -178,24 +178,15 @@ class DigitalOceanDomainRecordManager(DigitalOceanHelper, object):
     def __build_payload(self):
 
         payload = dict(
-            data=self.module.params.get("data"),
-            flags=self.module.params.get("flags"),
             name=self.module.params.get("name"),
-            port=self.module.params.get("port"),
-            priority=self.module.params.get("priority"),
             type=self.module.params.get("type"),
-            tag=self.module.params.get("tag"),
-            ttl=self.module.params.get("ttl"),
-            weight=self.module.params.get("weight"),
         )
 
-        # DigitalOcean stores every data in lowercase except TXT
-        if payload["type"] != "TXT" and payload["data"]:
-            payload["data"] = payload["data"].lower()
-
-        # digitalocean stores data: '@' if the data=domain
-        if payload["data"] == self.domain:
-            payload["data"] = "@"
+        payload_data = payload.get("data")
+        if payload_data:
+            # digitalocean stores data: '@' if the data=domain
+            if payload["data"] == self.domain:
+                payload["data"] = "@"
 
         return payload
 
--- a/plugins/modules/digital_ocean_domain_record_info.py
+++ b/plugins/modules/digital_ocean_domain_record_info.py
@@ -8,50 +8,46 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
-DOCUMENTATION = """
+DOCUMENTATION = r"""
 ---
 module: digital_ocean_domain_record_info
-author:
-  "Adam Papai (@woohgit)"
-  - Mark Mercado (@mamercad)
-version_added: 1.16.0
 short_description: Gather information about DigitalOcean domain records
 description:
-  - Gather information about DigitalOcean domain records
+  - Gather information about DigitalOcean domain records.
+version_added: 1.16.0
+author:
+  - "Adam Papai (@woohgit)"
+  - Mark Mercado (@mamercad)
 options:
   state:
     description:
-     - Indicate desired state of the target.
+      - Indicate desired state of the target.
     default: present
-    choices: [ present ]
+    choices: ["present"]
     type: str
-  record_id:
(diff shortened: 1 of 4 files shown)


plugins/modules/digital_ocean_domain_record_info.py
```diff
@@ -0,0 +1,227 @@
...
+            if status_code != 200:
+                self.module.exit_json(
+                    msg="Error getting domain records [%(status_code)s: %(json)s]"
+                    % {"status_code": status_code, "json": json}
+                )
+
+            for record in json["domain_records"]:
+                records.append(dict([(str(k), v) for k, v in record.items()]))
+
+            if "pages" in json["links"] and "next" in json["links"]["pages"]:
```
mpontillo, line 147: Potential `KeyError` on this line if `links` or `pages` don't exist in `json`. (Maybe prefer `.get()`.)

plugins/modules/digital_ocean_domain_record_info.py
```diff
@@ -0,0 +1,227 @@
...
+            status_code = response.status_code
+            json = response.json
+
+            if status_code != 200:
+                self.module.exit_json(
+                    msg="Error getting domain records [%(status_code)s: %(json)s]"
+                    % {"status_code": status_code, "json": json}
+                )
+
+            for record in json["domain_records"]:
```
mpontillo, line 144: Potential `KeyError` here if `domain_records` does not exist in `json`. (Maybe prefer `.get()`.)

plugins/modules/digital_ocean_domain_record_info.py
```diff
@@ -0,0 +1,227 @@
...
+            name=self.module.params.get("name"),
+            port=self.module.params.get("port"),
+            priority=self.module.params.get("priority"),
+            type=self.module.params.get("type"),
+            tag=self.module.params.get("tag"),
+            ttl=self.module.params.get("ttl"),
+            weight=self.module.params.get("weight"),
+        )
+
+        # DigitalOcean stores every data in lowercase except TXT
```
mpontillo, line 183: I don't really understand why we need to `.lower()` all the things here. I understand that `TXT` might seem like a special case, but I there are [a lot of other record types](https://en.wikipedia.org/wiki/List_of_DNS_record_types) and I feel like the general case should be "pass the user's input through unchanged".

Do you think it will cause a problem to _not_ call `.lower()`?

mamercad: The majority of this code was copied from the existing `digital_ocean_domain` module, including this section. With that in mind, I'd rather leave it alone (versus having a difference here and not there, so to speak). Though, if the comment ("DigitalOcean stores every data in lowercase except TXT") is true, I think that's actually okay since the [DNS specification](https://datatracker.ietf.org/doc/html/rfc4343) is "case insensitive".

plugins/modules/digital_ocean_domain_record_info.py
mamercad: See my comment about "this code is borrowed" below, heh.

mamercad: I've pushed changes.

plugins/modules/digital_ocean_domain_record_info.py
mamercad: I've pushed changes.

plugins/modules/digital_ocean_domain_record_info.py
mpontillo: The DNS specification is case insensitive in that the _name_ field is case insensitive (lookups for `digitalocean.com`, `DIGITALOCEAN.COM`, and `DiGiTAlOcEaN.CoM` will return identical results), but case might be important in non-`TXT` record types, and I'm hesitant to change user intent. For example, a `DNSKEY` resource record seems to very much be case sensitive, according to [RFC 4034](https://www.rfc-editor.org/rfc/rfc4034.html#section-2.3).

mpontillo: This may be a moot point, for now, because I don't know if there are non-`TXT` DNS records that the DO API supports that are, in fact, case sensitive. But since this is a new module, I feel like it might be better to leave this out?
On the other hand, why are we even sending in `data` here? This is an informational module, correct? (So it shouldn't be updating the `data` field, or mangling what comes back from the DO API?)

mamercad: I've pushed changes which remove the unused params, and, the `lower()`.

mpontillo (approved):
LGTM; you might want to add a test for getting "a lot of records" (to exercise pagination).

PR #238, opened 2022-04-27
Parameterize the DigitalOcean API base URL

### LINKED ISSUE #237: Allow overriding DigitalOcean API base URL
##### SUMMARY
Provide a way to override the official DigitalOcean API base URL.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
"DO API base URL override"

##### ADDITIONAL INFORMATION
Being able to override the official DO API base URL would selfishly facilitate using the DigitalOcean community collection within DigitalOcean, where we have test environments with their own unique base URLs.  This could also facilitate inspecting and debugging API requests and responses by overriding the base URL with one that specifies a proxy host/endpoint (as compared to instrumenting/debugging the Python source code).

```python
self.baseurl = os.getenv('DO_API_URL', 'https://api.digitalocean.com/')
```

It's not quite that easy as the URL definition is sprinkled throughout the various modules, but the main one being in class `DigitalOceanHelper`.

mamercad:
##### SUMMARY

Parameterize (and default) the DigitalOcean API base URL; fixes #237.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

* `digital_ocean`

### DIFF
--- a/changelogs/fragments/237-parameterize-do-api-baseurl.yaml
+++ b/changelogs/fragments/237-parameterize-do-api-baseurl.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - digital_ocean - parameterize the DigitalOcean API base url (https://github.com/ansible-collections/community.digitalocean/issues/237).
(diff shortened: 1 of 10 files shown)


mpontillo (approved):
lgtm! :ship:

PR #258, opened 2022-05-18
Added support for creating a load balancer with a tag

shuaibmunshi:
##### SUMMARY
Adding the ability to create a load balancer with a tag. Currently load balancers can only be created with a list of droplet IDs even though the API supports tags.

##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

digital_ocean_load_balancer

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/plugins/modules/digital_ocean_load_balancer.py
+++ b/plugins/modules/digital_ocean_load_balancer.py
@@ -780,6 +780,7 @@ def main():
         required_if=(
             [
                 ("state", "present", ["forwarding_rules"]),
+                ("state", "present", ["tag", "droplet_ids"], True),
             ]
         ),
         # Droplet ID and tag are mutually exclusive, check that both have not been defined
@@ -788,7 +789,6 @@ def main():
                 ("tag", "droplet_ids"),
             ]
         ),
-        required_one_of=[("tag", "droplet_ids")],
         supports_check_mode=True,
     )
 
--- a/plugins/modules/digital_ocean_load_balancer.py
+++ b/plugins/modules/digital_ocean_load_balancer.py
@@ -57,6 +57,15 @@ options:
     description:
       - An array containing the IDs of the Droplets assigned to the load balancer.
       - Required when creating load balancers.
+      - Mutually exclusive with tag, you can either define tag or droplet_ids but not both.
+    required: false
+    type: list
+    elements: int
+  tag:
+    description:
+      - A tag associated with the droplets that you want to dynamically assign to the load balancer.
+      - Required when creating load balancers.
+      - Mutually exclusive with droplet_ids, you can either define tag or droplet_ids but not both.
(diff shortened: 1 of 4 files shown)


shuaibmunshi:
@mamercad The sanity and unit tests should be fine now. 
I'm not quite sure what required_if is supposed to be doing but neither droplet_ids or tag are set to required because they are mutually exclusive so I removed both from required_if. If its supposed to be implemented some other way please let me know.

plugins/modules/digital_ocean_load_balancer.py
```diff
@@ -591,6 +606,12 @@ def create(self):
                     ),
                 )
 
+        # Droplet ID and tag are mutually exclusive, check that both have not been defined
+        if self.module.params.get('droplet_ids') is not None and self.module.params.get('tag') is not None:
```
mamercad, line 610: Using [`mutally_exclusive`](https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec-dependencies) for the parameter is probably a better way to go about this.

mamercad:
> @mamercad The sanity and unit tests should be fine now. I'm not quite sure what required_if is supposed to be doing but neither droplet_ids or tag are set to required because they are mutually exclusive so I removed both from required_if. If its supposed to be implemented some other way please let me know.

There's also a `required_one_of` which you can use ...

mamercad:
Once you get `black / lint` passing, it'd also be great if you could add integration tests which covers both cases (covering the failing cases would be great too).

mamercad:
Lastly -- thanks so much for working on this, we appreciate it! 🎉

plugins/modules/digital_ocean_load_balancer.py
shuaibmunshi: Thank you for pointing this out

shuaibmunshi:
@mamercad Sorry I forgot to test the case where tag and droplet_ids are not present and state is absent. I updated the code and ran the load_balancer integration tests myself, it should be fine now.

shuaibmunshi:
@mamercad unless I'm missing something I don't think the failed integration tests have anything to do with my changes

mamercad:
> @mamercad unless I'm missing something I don't think the failed integration tests have anything to do with my changes

They likely don't; I'll take a closer look when I get a bit of time -- sorry for the delay.

mamercad (changes requested):
Looks good, please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment).

mamercad:
I checked out this branch and ran the integration tests, they passed.

mamercad (approved):
Thanks for this contribution!

PR #278, opened 2022-07-12
Update `digital_ocean_load_balancer` (for `size_unit`)

### LINKED ISSUE #270: digital_ocean_load_balancer uses deprecated API calls
##### SUMMARY

I want to create a load balancer and allocate droplets to it via a 'tag'
I want to create a load balancer with a size of 3 nodes.

This module doesn't work with the current API spec:
https://docs.digitalocean.com/reference/api/api-reference/#operation/create_load_balancer

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

digital_ocean_load_balancer

##### COLLECTION VERSION

```paste below
community.digitalocean        1.19.0
```

mamercad:
##### SUMMARY

Add support for `size_unit` (versus `size`); deprecate `algorithm` -- fixes #270.
This field has been replaced by the `size_unit` field for all regions except in AMS2, NYC2, and SFO1.
The formula is lb-small = 1 node, lb-medium = 3 nodes, lb-large = 6 nodes.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- `digital_ocean_load_balancer`

### DIFF
--- a/plugins/modules/digital_ocean_load_balancer.py
+++ b/plugins/modules/digital_ocean_load_balancer.py
@@ -14,15 +14,9 @@ module: digital_ocean_load_balancer
 version_added: 1.10.0
 short_description: Manage DigitalOcean Load Balancers
 description:
-    - Manage DigitalOcean Load Balancers
+  - Manage DigitalOcean Load Balancers
 author: "Mark Mercado (@mamercad)"
 options:
-  oauth_token:
-    description:
-      - DigitalOcean OAuth token; can be specified in C(DO_API_KEY), C(DO_API_TOKEN), or C(DO_OAUTH_TOKEN) environment variables
-    type: str
-    aliases: ["API_TOKEN"]
-    required: true
   state:
     description:
       - The usual, C(present) to create, C(absent) to destroy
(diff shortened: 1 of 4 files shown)


plugins/modules/digital_ocean_load_balancer.py
```diff
@@ -415,13 +417,25 @@
 
 
 class DOLoadBalancer(object):
+
+    # Regions which use 'size' versus 'size_unit'
+    size_regions = ["ams2", "nyc2", "sfo1"]
```
mpontillo, line 422: Just a nit, but you _could_ use a `set` here:
```
size_regions = {"ams2", "nyc2", "sfo1"}

File: plugins/modules/digital_ocean_load_balancer.py [hunk 2]
```diff
@@ -415,13 +417,25 @@
...
+
     def __init__(self, module):
         self.rest = DigitalOceanHelper(module)
         self.module = module
         self.id = None
         self.name = self.module.params.get("name")
         self.region = self.module.params.get("region")
```
mpontillo, line 429: Is this region always going to come in as lowercase, or should you do something like `self.module.params.get("region").lower()`? (also, only do that if there is no possibility that it will come back as `None`!)

plugins/modules/digital_ocean_load_balancer.py
```diff
@@ -41,18 +35,24 @@
       - The available sizes are C(lb-small), C(lb-medium), or C(lb-large).
       - You can resize load balancers after creation up to once per hour.
       - You cannot resize a load balancer within the first hour of its creation.
+      - This field has been replaced by the size_unit field for all regions except in AMS2, NYC2, and SFO1.
```
mpontillo, line 38: Should we explain here what the user has to do? Something like:

> When creating a load balancer in the `AMS2`, `NYC2`, or `SFO1` regions, you must specify the `size` field. Otherwise, you must specify the `size_unit` field.

Should we also clarify that users can pass in both, to be on the safe side, and the module will figure out which one is needed based on the region?
I guess I'm also not clear on how error checking works. (are we not able to detect a missing field until we go to call the API?)

mpontillo (approved):

plugins/modules/digital_ocean_load_balancer.py
mamercad: Good call; will change.

plugins/modules/digital_ocean_load_balancer.py
mamercad: I'll make this impossible with `choices`, stay tuned.

plugins/modules/digital_ocean_load_balancer.py
mamercad: I think we're explaining it well enough in the relevant `description` items, and, I'm hoping this edge case goes away in time.

mamercad: We're making the `size` and `size_unit` parameters mutually-exclusive (Ansible will throw an error if they're both used) like this:

```python
        mutually_exclusive=(
            [
                ("tag", "droplet_ids"),
                ("size", "size_unit"),
            ]
        ),
```

And, we're using one-or-the-other like this:

```python
        if self.region in DOLoadBalancer.size_regions:
            self.module.params.pop("size_unit")
        else:
            self.module.params.pop("size")
```

I suppose we'll need to ensure that we have one-or-the-other after this pop -- I'll add a check for that.

mpontillo (approved):

PR #284, opened 2022-09-28
droplet: Allow creating droplet without droplet agent

coreywright:
##### SUMMARY

Allow passing `with_droplet_agent` boolean argument to droplet create API request to explicitly install or not the droplet agent which facilitates connecting to the droplet using the web console in the control panel.  The API default is to install the droplet agent.  This option allows a user to request not installing the droplet agent.

https://docs.digitalocean.com/reference/api/api-reference/#operation/droplets_create

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

digital_ocean_droplet

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/digital_ocean_droplet.py
+++ b/plugins/modules/digital_ocean_droplet.py
@@ -121,6 +121,14 @@ options:
     required: false
     default: false
     type: bool
+  with_droplet_agent:
+    description:
+      - Indicates whether to install the DigitalOcean agent used for providing access to the Droplet web console in the control panel.
+      - By default, the agent is installed on new Droplets but installation errors (i.e. OS not supported) are ignored.
+      - To prevent it from being installed, set to C(false).
+      - To make installation errors fatal, explicitly set it to C(true).
+    required: false
+    type: bool
   tags:
     description:
       - A list of tag names as strings to apply to the Droplet after it is created.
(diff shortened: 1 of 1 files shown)


mamercad (changes requested):
Looks good, needs a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) please.

coreywright:
well, it needs a changelog fragment, but more importantly (hopefully, eventually) some integration tests because of the uniqueness of the `with_droplet_agent` option, ie the API has interesting behavior that is only sometimes realized.

|API|supported image|agent installed|
|-|-|-|
|default (best-effort)|yes|yes|
|default (best-effort)|no|no|
|yes (required)|yes|yes|
|yes (required)|no|422|
|no|yes|no|
|no|no|no|

the only difference between "best-effort" and "required" is on an in-the-minority agent-unsupported image, eg `rancheros`, where "best-effort" quietly fails and "required" results in an API error (ie `422` with body `{"id":"unprocessable_entity","message":"The Droplet Agent is not supported with this configuration, please try an image with a different distribution."}`)

so integration tests would help insure that the API and module stay in sync as far as `with_droplet_agent` default behavior, but the key factor is the droplet image, but droplet images change regularly, complicating test configuration, longevity, and maintenance.

mamercad:
Closing as stale.

Issue #315, opened 2023-08-29
Release 1.24.0 broke using DO_API_TOKEN env variable


gizero:
##### SUMMARY

I'm using the dynamic inventory plugin. After upgrading the collection from 1.23.0 to 1.24.0 I got several CI pipelines failing with `HTTP Error 401: Unauthorized`. After rolling back to 1.23.0 everything went back to normal. All these scenarios expect the token to be provided via `DO_API_TOKEN` env variable, which is still supported as far as latest documentation suggests. Neither `oauth_token` nor `api_token` is set within the plugin configuration file.

To me it looks like https://github.com/ansible-collections/community.digitalocean/pull/301 somehow broke the expected behaviour of falling back to reading default env variables. Explicitly setting either `oauth_token` or `api_token` properties within config with something like `'{{ lookup("ansible.builtin.env", "DO_API_TOKEN") }}'` is an easy fix, but I believe this is unexpected and to be considered a bug.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

Dynamic inventory plugin

##### ANSIBLE VERSION

```paste below
# ansible --version
ansible [core 2.14.4]
  config file = /ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.10/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.10.6 (main, Mar 10 2023, 10:55:28) [GCC 11.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
# ansible-galaxy collection list community.digitalocean

# /root/.ansible/collections/ansible_collections
Collection             Version
---------------------- -------
community.digitalocean 1.24.0
```

##### CONFIGURATION

```paste below
# ansible-config dump --only-changed
CONFIG_FILE() = /ansible/ansible.cfg
DEFAULT_HOST_LIST(/ansible/ansible.cfg) = ['/ansible/inventory']
DEFAULT_REMOTE_USER(/ansible/ansible.cfg) = root
DEFAULT_ROLES_PATH(/ansible/ansible.cfg) = ['/root/.ansible/roles', '/ansible/roles']
HOST_KEY_CHECKING(/ansible/ansible.cfg) = False
[...]

PR #316, opened 2023-08-29
Inventory plugin: restore reading auth token from env variables

gizero:
##### SUMMARY
By completely removing any documentation for api_token / oauth_token, commit 609df3eaf506a5db13a960d5fc4bf4f70b5a3e50 suddenly broke support for reading the auth token from environment variables (which magically happens to work based on plugin options documentation).

Restore the documentation bits by matching docs fragment. While at it, also match the documented list of supported env variables which can be used as auth token sources.

Fixes #315

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Inventory plugin

PR #317, opened 2023-08-30
Release 1.24.1

mamercad:
Bugfixes
--------

- inventory plugin - restore reading auth token from env variables (https://github.com/ansible-collections/community.digitalocean/pull/315).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,14 @@ Community DigitalOcean Release Notes
 .. contents:: Topics
 
 
+v1.24.1
+=======
+
+Bugfixes
+--------
+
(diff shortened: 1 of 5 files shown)


mamercad:
@gizero We may have to go back to the drawing board, unfortunately. Also, this uncovered a gap in testing (not running the sanity and unit tests on #316.

gizero:
😮 I can reproduce the sanity test failure locally... Strangely enough the change is working. All listed envs work as expected. I'll try to dig deeper to figure out why the test is complaining.

raman-babich:
@mamercad looks like we can close this one, doesn't it?

